PR #10, opened 2020-02-11
Update Source Code diagram

ionut-arm:
This commit updates the source code diagram and the draw.io canvas associated
with it.

image


hug-dev:
Also, you could put the tss-esapi crate linked to the TPM provider if you wanted!

ionut-arm:
> Thanks for the updated changes! It will help the reader 😀
> Some comments:
> 
>     * `service_back` and `service_front` should maybe be renamed `back` and `front` to align to what we have
> 
>     * maybe instead of the `protobuf` block we can have the `operations` crate (and inside it the `protobuf` folder)? Just to be clear that any language-agnostic format could be used, and communicate on that `operations` crate.
> 
>     * would be nice to update the crate names with the actual ones (`parsec`, `parsec-interface-rs`, `parsec-operations`).
> 
>     * what is the role of the most top-level `interface` block? Maybe that could be remove to add more clarity? Things inside would be kept obviously.

Pushed changes to fix most of those issues. Didn't make `parsec-operations` look like a crate because it isn't one! But otherwise everything was updated.

PR #21, opened 2020-03-24
Update operations and response codes with PSA 1.0

hug-dev:
Updates the operation documentation to match the update on our protobuf
oeprations.
Also update the response code status and the generate key operation.


src/parsec_client/operations/psa_algorithm.md
```diff
@@ -0,0 +1,347 @@
...
+key.
+
+*Notice:* not all possible attributes are currently supported by Parsec. Please see the [API
+coverage](../../api_coverage.md) for an overview of what Parsec currently supports.
+
+## Algorithm message
+
+| Field name | Type    | Description                                 |
+|------------|---------|---------------------------------------------|
+| `variant`  | `oneof` | Key type, see below for the possible values |
```
ionut-arm: I'm unsure whether we should be making this protobuf-specific, since theoretically that could be replaced with anything.

src/parsec_client/operations/psa_algorithm.md
```diff
@@ -0,0 +1,347 @@
...
+The key can still be used for non-cryptographic actions such as exporting, if permitted by the usage
+flags.
+
+#### Hash enumeration
+
+Possible hash algorithms.
+
+- `HASH_NONE`: this default variant should not be used.
+- `MD2`: MD2. Deprecated: the MD2 hash is weak and deprecated and is only recommended for use in
+   legacy protocols.
```
ionut-arm: Do you think it's worth mentioning that it will be at the discretion of the system admin whether legacy algorithms are permitted?

src/parsec_client/operations/psa_algorithm.md
```diff
@@ -0,0 +1,347 @@
...
+   128 bits (16 bytes).
+- `CFB`: A stream cipher built using the Cipher Feedback (CFB) mode of a block cipher. The
+   underlying block cipher is determined by the key type.
+- `OFB`: A stream cipher built using the Output Feedback (OFB) mode of a block cipher. The
+   underlying block cipher is determined by the key type.
+- `XTS`: The XTS cipher mode of a block cipher. XTS is a cipher mode which is built from a block
+   cipher. It requires at least one full block of input, but beyond this minimum the input does not
+   need to be a whole number of blocks.
+- `ECB_NO_PADDING`: The Electronic Code Book (ECB) mode of a block cipher, with no padding. The
+   underlying block cipher is determined by the key type. **Warning:** ECB mode does not protect the
```
ionut-arm: Just to be consistent with the other uses of `**Warning**`, should that part be on a new line? Or is it like this because it's in a list and it would make it look weird?

src/parsec_client/operations/psa_algorithm.md
```diff
@@ -0,0 +1,347 @@
...
+Asymmetric signature algorithms.
+
+| Field name | Type    | Description                                                       |
+|------------|---------|-------------------------------------------------------------------|
+| `variant`  | `oneof` | Asymmetric signature algorithm, see below for the possible values |
+
+| `variant` oneof field name | Type                    |
+|----------------------------|-------------------------|
+| `rsa_pkcs1v15_sign`        | `RsaPkcs1v15Sign`       |
+| `RsaPkcs1v15SignRaw`       | `rsa_pkcs1v15_sign_raw` |
```
ionut-arm: Something about the naming tells me these should've been the other way around

src/parsec_client/operations/psa_algorithm.md
```diff
@@ -0,0 +1,347 @@
...
+RSA PKCS#1 v1.5 signature with hashing. This is the signature scheme defined by RFC 8017 (PKCS#1:
+RSA Cryptography Specifications) under the name RSASSA-PKCS1-v1_5.
+
+| Field name | Type   | Description           |
+|------------|--------|-----------------------|
+| `hash_alg` | `Hash` | Hash algorithm to use |
+
+##### RsaPkcs1v15SignRaw message
+
+Raw PKCS#1 v1.5 signature. The input to this algorithm is the DigestInfo structure used by RFC 8017
```
ionut-arm: You might hate me for suggesting this, but - think it's worth adding links to the various RFC (sections) mentioned around here?

src/parsec_client/operations/psa_algorithm.md
hug-dev: I was thinking about that as well! Was unsure if, in the book, the operation documentation should be common to all body types or not.

If it is, I should modify this PR to make it totally generic: do not assume any types or any name. Basically take the PSA Crypto API specification and translate it into a generic Parsec version, language agnostic. Then we would have to document how from this generic model, the different things are implemented in all body type languages like protobuf and others.

If we choose to make it body type language specific, then we would have to add one more level on the book below the "Operations" section for each body type language. That would mean some repetition in the documentation.

src/parsec_client/operations/psa_algorithm.md
hug-dev: Yes good point! Will add it in the overview page.

src/parsec_client/operations/psa_algorithm.md
hug-dev: I think it's the later, you can not have list items on multiple lines AFAIK

src/parsec_client/operations/psa_algorithm.md
hug-dev: Good catch! My `vim` macro failed me there

src/parsec_client/operations/psa_algorithm.md
hug-dev: Yep, will do!

src/parsec_client/operations/psa_algorithm.md
ionut-arm: Not sure if we can decide that for ourselves!

@paulhowardarm - any thoughts on this? Should these docs (the ones in the operations folder) be specific to protobuf?

paulhowardarm: The original idea was for the operation directory to be neutral, and for the "Contract" section at the bottom to provide the links to interface definitions such as protobuf. So the expected developer flow was that you'd start off with the markdown document to get the general overview of the operation, and a neutral description of the ins/outs, you'd then jump the hyperlink to the (say) protobuf file to see the contract and any additional docs on a per-field or per-parameter basis.

Now there are problems with this original view. The first problem is that documentation looks far better in markdown than it does in protobuf where it's just unformatted text. I thought that we could solve this by auto-generating MD docs from the protobuf files, but then you need a build step to generate the docs and you can't just read it in GitHub. I think it's important that our documentation should be readable from the source without requiring a build, so auto-generation doesn't work. The second problem is that you would need to replicate all that stuff when a new serialisation format comes along (if indeed that ever happens - there's no immediate prospect of that).

If all our API doc is to be static MD with no auto-generation step, then I think the gold standard would be for all of the doc to be format-neutral, such that you can read it and then jump the hyperlink to the protobuf (or whatever) file, and the content in that file should be self-evident from what you've been reading, without any significant extra doc in that file. This doesn't have to mean that the doc was protobuf-specific, but it should have been descriptive enough to make the protobuf understandable by itself.

The question is whether that can be achieved in practice. If you look at what's there today, some of our existing content follows that ideal. For example, the **PsaCreateKey** MD file has a "**Parameters**" section with `key_name` and `key_attributes`. And when you jump to the protobuf file, you see a protobuf `Operation` message with fields `key_name` and `attributes` and the precise data types. In my head, that's very clear and self-evident, and yet we've managed to keep the MD document format-neutral.

So the question is: can we do that everywhere? Can we explain key attributes in a way that explains what it means, and what all the options are, but then link to the protobuf file in such a way that it simply confirms what we've already written, or will contain content that exactly matches what the reader will expect to see?

If we can do that, then I think it's the winning solution.

If we can't, then probably the next best thing is to change the hyperlinks so that "Protobuf" doesn't go straight to the protobuf file itself, but instead goes to an interim MD file that explains whatever part of the contract absolutely requires a protobuf-specific fragment of doc, and then have that interim MD file have another hyperlink that goes to the protobuf message file itself. A level of indirection, in other words. But we only do it in a few well-judged places, rather than structure the entire book in this way.

Does that make sense?

hug-dev: Thanks for the detailed explanation! That makes a lot of sense. Having format-neutral documentation is the best thing to do to avoid any kind of duplication and to have more readable documentation as well.
I think for most operations/structures that we describe in a format-neutral way from the PSA Crypto API, the translation and understanding of the Protobuf format will be implicit.

Will modify this PR then, to put it back to be formal-neutral!

hug-dev:
Hey @ionut-arm!
Just upated, addressing all the comments above. I made a new commit so you can see the specific changes.
I removed the protobuf-specific parts.

When it is needed for a type to be referred amoung multiple locations like in the parameters of an operation I wrote this type in fixed-width font like `RsaKeyPair` for example.
When it is not needed for the type to be referenced/defined, I just used a human-understandable language like "CBC with PKCS#7 padding" which was previously the protobuf value `CBC_PKCS7`. That way, it gives freedom to implementors.

src/parsec_client/operations/psa_key_attributes.md
```diff
@@ -3,206 +3,191 @@
 The attributes are used to fully describe a cryptographic key: its type, size and what is permitted
 to do with that key.
 
-*Notice:* not all possible attributes are currently supported by Parsec. Please see the [API
+Some of the algorithms defined here are deprecated and should not be used without a valid reason. It
+is at the discretion of the system administrator whether those algorithms are permitted or not.
+
+**Notice:** not all possible attributes are currently supported by Parsec. Please see the [API
 coverage](../../api_coverage.md) for an overview of what Parsec currently supports.
```
ionut-arm: Do you think it's also worth saying that some of the algorithms/parameters might not be supported by some providers at all, because it's not in their interface? I'm not sure if that's the case with PKCS11 or TPM, but maybe in the future something like this would happen

hug-dev: Done! I also squashed all commits together.

src/parsec_client/operations/psa_key_attributes.md
```diff
@@ -0,0 +1,204 @@
...
+| `key_algorithm`   | [`Algorithm`](psa_algorithm.md#algorithm-type) | Permitted algorithms to be used with the key |
+
+## UsageFlags type
+
+Definition of the usage flags. They encode what kind of operations are permitted on the key. A
+`UsageFlags` type contains the following members:
+
+| Name             | Type   | Description                                        |
+|------------------|--------|----------------------------------------------------|
+| `export`         | `bool` | Permission to export the key                       |
```
ionut-arm: Nitpick: "`bool`" should probably be "boolean", since it shouldn't be a type but a word like "Unsigned integer" on line 20

hug-dev: Good point, will keep that change on my local changes and add it to the next update of the book.

PR #23, opened 2020-04-03
Update content and format of all operations

hug-dev:
Updates all operations to have a standard format and updates the PSA
operations to match exactly the 1.0.0 specs.


src/parsec_client/operations/list_providers.md
```diff
@@ -1,10 +1,36 @@
...
+|-------------|------------------------------------------------|------------------------------|
+| `providers` | Vector of [`ProviderInfo`](#providerinfo-type) | List of provider information |
+
+### ProviderInfo type
+
+A `ProviderInfo` type contains the following members:
+
+| Name          | Type             | Description                                                               |
+|---------------|------------------|---------------------------------------------------------------------------|
+| `uuid`        | String           | Unique, permanent, identifier of the provider                             |
```
ionut-arm, line 21: Mind adding that this is a [version 4 UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random))?

src/parsec_client/operations/list_providers.md
```diff
@@ -1,10 +1,36 @@
...
+| `description` | String           | Short description of the provider                                         |
+| `vendor`      | String           | Provider vendor                                                           |
+| `version_maj` | Unsigned integer | Provider implementation version major                                     |
+| `version_min` | Unsigned integer | Provider implementation version minor                                     |
+| `version_rev` | Unsigned integer | Provider implementation version revision number                           |
+| `id`          | Unsigned integer | Provider ID to use on the wire protocol to communicate with this provider |
```
ionut-arm, line 27: Are we still certain that we want to support at most one provider of each type, per service?

src/parsec_client/operations/psa_export_public_key.md
```diff
@@ -1,41 +1,65 @@
...
-## Result values
+The `RSAPublicKey` representation is:
 
-- **key_data** - Bytes of the key in one of the formats described above.
+```
+RSAPublicKey ::= SEQUENCE {
+    modulus INTEGER, -- n
+    publicExponent INTEGER -- e
+}
+```
```
ionut-arm, line 62: Could you mark this section somewhow? Maybe like in books, by putting a [superscript ](https://www.fileformat.info/info/unicode/char/00b9/index.htm) where it's mentioned above and then the same one here, before "The `RsaPublicKey`..."

src/parsec_client/operations/psa_sign_hash.md
```diff
@@ -0,0 +1,45 @@
...
+## Description
+
+With most signature mechanisms that follow the hash-and-sign paradigm, the hash input to this
+function is the hash of the message to sign. The hash algorithm is encoded in the signature
+algorithm. Some hash-and-sign mechanisms apply a padding or encoding to the hash. In such cases, the
+encoded hash must be passed to this function. The current version of this specification defines one
+such signature algorithm: Raw PKCS#1 v1.5 signature.
+
+**Note:** To perform a hash-and-sign algorithm, the hash must be calculated before passing it to
+this function. This can be done with the operation PsaHashCompute or with a multi-part hash
```
ionut-arm, line 38: I would remove the reference to `PsaHashCompute` until we actually have that operation 😬

src/parsec_client/operations/psa_verify_hash.md
```diff
@@ -0,0 +1,47 @@
...
+## Description
+
+With most signature mechanisms that follow the hash-and-sign paradigm, the hash input to this
+function is the hash of the message to sign. The hash algorithm is encoded in the signature
+algorithm. Some hash-and-sign mechanisms apply a padding or encoding to the hash. In such cases, the
+encoded hash must be passed to this function. The current version of this specification defines one
+such signature algorithm: Raw PKCS#1 v1.5 signature.
+
+**Note:** To perform a hash-and-sign verification algorithm, the hash must be calculated before
+passing it to this function. This can be done by calling PsaHashCompute or with a multi-part hash
```
ionut-arm, line 39: Same here

src/parsec_client/operations/list_providers.md
hug-dev: I think this spec does not prevent it: currently our version of a "Provider ID" is a fixed number in an enum but we could expand it later to be dynamically allocated number and this spec won't change. I think that even if this number is dynamically allocated, it will still be called "Provider ID".

src/parsec_client/operations/psa_export_public_key.md
hug-dev: I achieved that with `[1]` style references as superscript in Markdown is done via HTML tags which are not supported by our Markdown format :/

src/parsec_client/operations/psa_sign_hash.md
hug-dev: Well even so, I think it is correct to say that `PsaHashCompute` will be intended to cover this, even if it is not implemented right now. I think the fact that there is no link to the operation page is enough for the user to understand that it is not currently implemented 😋

src/parsec_client/operations/list_providers.md
ionut-arm: True, I guess I was thinking of both UUID and ID at the same time!

src/parsec_client/operations/psa_sign_hash.md
ionut-arm: > I think the fact that there is no link to the operation page is enough for the user to understand that it is not currently implemented 

Not if you say "This **can** be done...". Maybe change to "could" or something else, but "can" strictly implies that it can be done as of now.

ionut-arm: (but even if you change it, I, for one, would be frustrated/confused if I read in the documentation of a project that I can use feature X when that feature clearly doesn't exist)

hug-dev: Fair enough, I will add "could" and clearly write that the operation is not yet implemented.

src/parsec_client/operations/psa_export_public_key.md
ionut-arm: Nice! `[1]` works - the "superscript" I linked to above is just a UTF-8 character so that wouldn't have broken the MD format, but `[1]` is probs better

PR #27, opened 2020-04-20
Rename key ID to key info

hug-dev:
The Key ID Manager has been renamed Key Info Manager after the
attributes were added into what is stored.

See parallaxsecond/parsec#147 for reference.


src/parsec_service/key_info_managers.md
```diff
@@ -0,0 +1,13 @@
+# Key Info Managers
+
+The key info stored by the key info managers consist in a structure composed of: * an `id`, which is
+the reference to a key in the specific provider * `attributes`, which are the attributes of the key
```
ionut-arm, line 4: Is that supposed to be/look like a list?

hug-dev: Oops! Forgot a space, nice catch

src/parsec_service/source_code_structure.md
```diff
@@ -86,12 +86,12 @@ underlying security hardware to determine if the request has been correctly auth
...
-allows keys to be addressed by a UTF-8 name, and the providers which have specific requirements for
-key handles (e.g. Mbed Crypto uses a 32 bit value for handles of persistent keys). Using a key ID
-manager is only required for persistent keys and the only current implementation stores the mappings
-on disk.
+The `key_info_managers` module offers functionality for the providers for persisting mappings
+between key names and provider-specific key information. The key info managers close the gap between
+the API, which allows keys to be addressed by a UTF-8 name, and the providers which have specific
+requirements for key handles (e.g. Mbed Crypto uses a 32 bit value for handles of persistent keys).
+Using a key info manager is only required for persistent keys and the only current implementation
+stores the mappings on disk.
```
ionut-arm, line 94: While this is certainly useful to know, I'm not sure it should be here. It feels like we have a lot of information scattered over various documents and it'll be difficult to make sure we keep everything updated - e.g. for this phrase, I'm not sure we'll remember to update it when we'll have a different key info manager, since this page seems only vaguely related to what managers we actually have. We can create an issue to track this, potentially (re)moving doc parts that aren't in the most obvious place.

hug-dev: True, will create one.

PR #30, opened 2020-05-06
Make guides up-to-date

hug-dev:
Specifies the clients and their current state. Makes the guides
up-to-date and easier to reproduce.


hug-dev:
https://crates.io/crates/parsec-client has been found to be a dead link by the CI but it is definitely working. I think the CI can be ignored on this one.

PR #34, opened 2020-06-16
Add info about building TPM/PKCS11 providers

hug-dev:

### DIFF
--- a/src/parsec_client/operations/psa_algorithm.md
+++ b/src/parsec_client/operations/psa_algorithm.md
@@ -52,9 +52,6 @@ Possible hash algorithms are:
 - **SHA3-256.**
 - **SHA3-384.**
 - **SHA3-512.**
-- **Any Hash.** In a hash-and-sign algorithm policy, allow any hash algorithm. This value must not
-   be used to build an algorithm specification to perform an operation. It is only valid to build
-   policies.
 
 ### Mac algorithm
 
@@ -146,6 +143,11 @@ Asymmetric signature algorithms. Supported algorithms:
 - **Deterministic ECDSA signature with hashing.** This is the deterministic ECDSA signature scheme
    defined by [RFC 6979](https://tools.ietf.org/html/rfc6979.html). Uses one of the hash algorithm
    supported.
+
+When defining the permitted algorithms in a key policy, the hash-and-sign algorithms above can use
+the value **Any Hash** for their hash algorithm, meaning that it will allow any hash algorithm. This
+value must not be used to build an algorithm specification to perform an operation. It is only valid
+to build policies.
 
 ### AsymmetricEncryption algorithm
 
--- a/src/parsec_service/build_run.md
+++ b/src/parsec_service/build_run.md
@@ -16,7 +16,9 @@ example of the ones used for testing, for example [the TPM provider
 one](https://github.com/parallaxsecond/parsec/blob/master/e2e_tests/provider_cfg/tpm/config.toml).
 
 The `mbed-crypto-provider` feature is going to be used as an example in this guide. This can be
-replaced by a subset of the features mentioned above, space or comma separated.
+replaced by a subset of the features mentioned above, space or comma separated. If you would like to
+test the TPM or PKCS check the [related
+guides](test.md#testing-the-tpm-provider-using-the-software-tpm).
 
 To build and run Parsec from source:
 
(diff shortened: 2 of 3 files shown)


src/parsec_client/operations/psa_algorithm.md
```diff
@@ -147,6 +144,11 @@ Asymmetric signature algorithms. Supported algorithms:
    defined by [RFC 6979](https://tools.ietf.org/html/rfc6979.html). Uses one of the hash algorithm
    supported.
 
+When defining the permitted algorithms in a key policy, the hash-and-sign algorithms above can use
+the value **Any Hash** for their hash algorithm, meaning that it will allow any hash algorithm. This
+value must not be used to build an algorithm specification to perform an operation. It is only valid
+to build policies.
```
ionut-arm, line 150: I totally agree with this, but I'd like to enforce it somewhere before the providers as well - in the interface or client, maybe. Not compulsory, just thinking out loud here

src/parsec_service/build_run.md
```diff
@@ -74,14 +76,20 @@ the name `protobuf-compiler`:
 sudo apt install protobuf-compiler
 ```````
 
+To use `bindgen` and generate the Rust to C wrappers, `libclang` (version at least 3.9) is needed:
+
+```````
+sudo apt install llvm-dev libclang-dev clang
```
ionut-arm, line 82: This is for Debian-based distros - do we want to specify the command for other distributions/for MacOS too?

src/parsec_service/test.md
```diff
@@ -100,4 +100,75 @@ target and then observing the code segments reached using said input. To stop th
...
+
+If you do not have a hardware TPM, or you do not want to use it for tests, you might want to use the
+IBM Software TPM. You will need to follow these steps:
+
+- install the TSS libraries ([installation
+   guide](https://github.com/tpm2-software/tpm2-tss/blob/master/INSTALL.md)). Parsec has been tested
+   with version 2.3.3.
+- install the [TPM tools](https://github.com/tpm2-software/tpm2-tools)
+- install the [Software
+   TPM](https://sourceforge.net/projects/ibmswtpm2/files/ibmtpm1332.tar.gz/download), run `make` in
```
ionut-arm, line 113: Is it possible to point that URL to a more generic link? I'm afraid of forgetting to ever update it since they do sometimes push new versions out.

src/parsec_service/test.md
```diff
@@ -100,4 +100,75 @@ target and then observing the code segments reached using said input. To stop th
...
+
+Use the TPM tools to startup the TPM server and set its owner hierarchy password (the tools need to
+be on the `PATH`):
+
+```
+$ tpm2_startup -c -T mssim
+$ tpm2_changeauth -c owner tpm_pass -T mssim
+```
+
+Start Parsec with the TPM configuration:
```
ionut-arm, line 134: 💯 Thanks, this will save the sanity of a few people out there

src/parsec_client/operations/psa_algorithm.md
hug-dev: Aren't we doing that with the `SignHash` type?

src/parsec_service/build_run.md
hug-dev: I think ideally yes, but I guess most people know how to translate the command to the ones of their own distro, assuming the package names do not change.

src/parsec_client/operations/psa_algorithm.md
ionut-arm: I meant the fact that policy algorithms support wildcards, but operation algorithms don't, we don't enforce that do we?

src/parsec_service/test.md
hug-dev: Sure! I will change it to something more generic. On the same topic, there is now a new software TPM used by default by the stack that we use: https://github.com/tpm2-software/tpm2-tss#software-tpm

src/parsec_client/operations/psa_algorithm.md
hug-dev: I think we do because `Any` has been removed from [`Hash`](https://docs.rs/parsec-interface/0.15.0/parsec_interface/operations/psa_algorithm/enum.Hash.html)

ionut-arm: ? [`PsaSignHash`](https://github.com/parallaxsecond/parsec-interface-rs/blob/master/src/operations/psa_sign_hash.rs#L18) takes an `AsymmetricSignature` algorithm. [`AsymmetricSignature`](https://github.com/parallaxsecond/rust-psa-crypto/blob/master/psa-crypto/src/types/algorithm.rs#L317) accepts `SignHash` as the hash algorithm. So there's nothing preventing a client using `Any` for `PsaSignHash`

hug-dev: oh absolutely, my bad 😅

PR #39, opened 2020-06-24
Add consequences of unmitigations

hug-dev:


src/threat_model/threat_model.md
```diff
@@ -301,14 +301,14 @@ cleared.
...
+| 0  | An attacker can use all the threads of the service's thread pool, requesting for a time-consuming operation on each one of them, to leave the less amount of free time possible to other clients.                                                                                                                                                                                                                                                                            | The requests take more time to execute.                                                                              |
+| 1  | Physical Denial of Service attacks are unmitigated.                                                                                                                                                                                                                                                                                                                                                                                                                          | The Parsec service is stopped.                                                                                       |
```
ionut-arm, line 307: I wouldn't say a physical denial of service attack means the service is stopped, it will just return continuous errors.

And actually, we don't mitigate ANY physical attacks - i.e. if someone comes and replaces the normal HSM with another one, then takes it away full of keys.

ionut-arm: Or do we have an assumption about that?

hug-dev: > I wouldn't say a physical denial of service attack means the service is stopped, it will just return continuous errors.

I guess I was mentioning the worst case, if someone actually throws the platform Parsec is running in the water or something similar.

> Or do we have an assumption about that?

Yes, we say in [ASUM-1](https://parallaxsecond.github.io/parsec-book/threat_model/threat_model.html#assumptions) that the hardware modules are physically secure.

ionut-arm: Wait, so if we have an assumption why do we have an unmitigation too o_O

hug-dev: YEs, true and it is referred only once, can delete

PR #42, opened 2020-06-29
Added asymmetric encrypt and decrypt operation pages.

sbailey-arm:

### DIFF
--- a/src/parsec_client/operations/psa_asymmetric_decrypt.md
+++ b/src/parsec_client/operations/psa_asymmetric_decrypt.md
@@ -1,0 +1,39 @@
+# PsaAsymmetricDecrypt
+
+Decrypt a short message with a private key. Opcode: 9 (`0x0009`)
+
+## Parameters
+
+| Name         | Type                                                                      | Description                                                                |
+|--------------|---------------------------------------------------------------------------|----------------------------------------------------------------------------|
+| `key_name`   | String                                                                    | Name of the key to use for the operation                                   |
+| `alg`        | [`AsymmetricEncryption`](psa_algorithm.md#asymmetricencryption-algorithm) | An asymmetric encryption algorithm that is compatible with the type of key |
+| `ciphertext` | Vector of bytes                                                           | Short message to decrypt                                                   |
+| `salt`       | Vector of bytes                                                           | Salt to use during encryption, if supported by the algorithm               |
+
+- `key_name` must be the name of an RSA asymmetric key pair. The key must allow the [usage
+   flag](psa_key_attributes.md#usageflags-type) `decrypt`.
+- `salt` can be provided if supported by the algorithm. If the algorithm does not support salt, pass
+   an empty vector. If the algorithm supports optional salt, pass an empty vector to indicate no
+   salt. For RSA PKCS#1 v1.5 encryption, no salt is supported.
+
+## Results
+
+| Name        | Type            | Description                             |
+|-------------|-----------------|-----------------------------------------|
+| `plaintext` | Vector of bytes | Buffer containing the decrypted message |
+
+## Specific response status codes
+
+- `PsaErrorNotPermitted`: The key does not have the `decrypt` flag, or it does not permit the
+   requested algorithm.
+
+## Description
+
+This function will decrypt a short message with the private key of the provided key pair.
+
+## Contract
+
+[Protobuf](https://github.com/parallaxsecond/parsec-operations/blob/master/protobuf/psa_asymmetric_decrypt.proto)
+
+*Copyright 2020 Contributors to the Parsec project.*
--- a/src/api_coverage.md
+++ b/src/api_coverage.md
@@ -22,32 +22,32 @@ in the future and will be organized by operation type.
 
 ### Single part operations
 
-| Operation name                     | Protobuf contract exists                                                                                    |
-|------------------------------------|-------------------------------------------------------------------------------------------------------------|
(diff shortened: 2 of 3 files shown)


src/parsec_client/operations/psa_asymmetric_decrypt.md
```diff
@@ -0,0 +1,38 @@
...
+| `key_name` | String                                                                    | Name of the key to use for the operation                                   |
+| `alg`      | [`AsymmetricEncryption`](psa_algorithm.md#asymmetricencryption-algorithm) | An asymmetric encryption algorithm that is compatible with the type of key |
+| `message`  | Vector of bytes                                                           | Short message to decrypt                                                   |
+| `salt`     | Vector of bytes                                                           | Salt to use during encryption, if supported by the algorithm     |
+
+- `key_name` must be the name of an RSA asymmetric key pair or private key. The key must allow the [usage
+   flag](psa_key_attributes.md#usageflags-type) `decrypt`.
+- `salt` can be provided if supported by the algorithm. If the algorithm does not support salt, pass NULL. If the 
+algorithm supports optional salt, pass NULL to indicate no salt. For RSA PKCS#1 v1.5 encryption, no salt is supported.
```
ionut-arm, line 17: The references to `NULL` should probably be replaced with `empty vector` or something like that.

src/parsec_client/operations/psa_asymmetric_decrypt.md
```diff
@@ -0,0 +1,38 @@
...
+## Parameters
+
+| Name       | Type                                                                      | Description                                                                                                           |
+|------------|---------------------------------------------------------------------------|----------------------------------------------------------------------------|
+| `key_name` | String                                                                    | Name of the key to use for the operation                                   |
+| `alg`      | [`AsymmetricEncryption`](psa_algorithm.md#asymmetricencryption-algorithm) | An asymmetric encryption algorithm that is compatible with the type of key |
+| `message`  | Vector of bytes                                                           | Short message to decrypt                                                   |
+| `salt`     | Vector of bytes                                                           | Salt to use during encryption, if supported by the algorithm     |
+
+- `key_name` must be the name of an RSA asymmetric key pair or private key. The key must allow the [usage
```
hug-dev: > or private key

I think you can not have RSA private key on their own

src/parsec_client/operations/psa_asymmetric_decrypt.md
hug-dev: or nothing at all (as in remove the sentence)? Because I guess it will depend how the optional field is implemented in the target language. For example in Rust, it is `None` I think.

src/parsec_client/operations/psa_asymmetric_decrypt.md
```diff
@@ -0,0 +1,40 @@
...
+
+Decrypt a short message with a private key. Opcode: 9 (`0x0009`)
+
+## Parameters
+
+| Name       | Type                                                                      | Description                                                                |
+|------------|---------------------------------------------------------------------------|----------------------------------------------------------------------------|
+| `key_name` | String                                                                    | Name of the key to use for the operation                                   |
+| `alg`      | [`AsymmetricEncryption`](psa_algorithm.md#asymmetricencryption-algorithm) | An asymmetric encryption algorithm that is compatible with the type of key |
+| `message`  | Vector of bytes                                                           | Short message to decrypt                                                   |
```
hug-dev, line 11: For decrypt we have `message` in the parameters and `decrypted_message` in the results and for encrypt it is `message` and `encrypted_message`. I propose to go for fully descriptive name and use `decrypted_message` and `encrypted_message` only OR use `message` only and let the consumer be smart enough to see that depending on the operation and the parameters/results context what is the type of the message. I think the first option is more friendly 👍

hug-dev:
The CI will fail because the links will not work yet (as the files on the operations repository do not yet exist). I propose that, once this PR has been approved, you go ahead and implement it in protobuf on the operations repository and once done, we can restart the CI here and merge this.

sbailey-arm:
> 
> 
> The CI will fail because the links will not work yet (as the files on the operations repository do not yet exist). I propose that, once this PR has been approved, you go ahead and implement it in protobuf on the operations repository and once done, we can restart the CI here and merge this.

Ok, I have made some changes as per the comments, will get protobuf ready

src/parsec_client/operations/psa_asymmetric_decrypt.md
ionut-arm: How about `ciphertext` and `plaintext`?

ionut-arm: Although since this has already been resolved, it's probably ok as it is!

src/parsec_client/operations/psa_asymmetric_decrypt.md
ionut-arm: It also depends on the contract type - it might be like this in protobuf, maybe CBOR would be different. I said empty vector because the field is declared as vector above

src/parsec_client/operations/psa_asymmetric_decrypt.md
sbailey-arm: I've changed to Ionut's suggestion to keep it the same as `parsec-operations`.

hug-dev:
Restarting CI

ionut-arm:
Could you also update the [API coverage](https://parallaxsecond.github.io/parsec-book/api_coverage.html#single-part-operations) page? Either here or in a different PR. Forgot about that when reviewing...

PR #46, opened 2020-07-09
How-to for adding operations to Parsec with the mbed-crypto backend

sbailey-arm:


src/contributing/adding_new_operation_how_to.md
```diff
@@ -0,0 +1,108 @@
...
+
+This mini how-to will cover how to add operations to the backend and frontend of Parsec. The steps
+for adding operations to a backend will differ between providers, this guide will cover Mbed Crypto.
+The steps for adding operations to the frontend should be the same for all backend providers.
+
+## Backend (Mbed Crypto)
+
+### psa-crypto-sys
+
+Locate the parts of the PSA API in `mbedtls` that will need to be used during the operation you are
```
hug-dev, line 11: I think you can replace `mbedtls` with "Mbed Crypto" as it is the actual component implementing the PSA APIs, even if contained in the Mbed TLS project.

src/contributing/adding_new_operation_how_to.md
```diff
@@ -0,0 +1,108 @@
+# Adding operations to Parsec
+
+This mini how-to will cover how to add operations to the backend and frontend of Parsec. The steps
```
hug-dev, line 3: "Backend" and "Frontend" have a different meaning, than what they usually use to describe, in Parsec. See the [repository map](https://parallaxsecond.github.io/parsec-book/parsec_service/source_code_structure.html#repository-map) diagram. `Frontend` parts are the listeners + front end handler and "backend" parts are the dispatcher and back end handler. I think to not confuse anybody we should probably use other terms throughout this document.

> This mini how-to will cover how to add operations to the backend and frontend of Parsec.

Can maybe be replaced with: "This mini how-to will cover how to add operations to the Parsec service and Rust client."

We should also maybe put a note that the `parsec-interface-rs` and client parts are only for the Rust client but should be similar for other clients.

src/contributing/adding_new_operation_how_to.md
```diff
@@ -0,0 +1,108 @@
...
+for adding operations to a backend will differ between providers, this guide will cover Mbed Crypto.
+The steps for adding operations to the frontend should be the same for all backend providers.
+
+## Backend (Mbed Crypto)
```
hug-dev, line 7: For the section names, here is what I can propose:

```
## Operation specification
### parsec-book
### parsec-operations

## Parsec Rust Interface
With `parsec-interface-rs`

## Parsec Rust Client

## Parsec Service
### psa-crypto-sys
### psa-crypto
### parsec
``` 

What do you think?

src/contributing/adding_new_operation_how_to.md
```diff
@@ -0,0 +1,108 @@
...
+that the Rust interface handles all possible situations (e.g. closing a key-handle on an error) to
+ensure it is a safe Rust native interface to the operation.
+
+At this point, you will now have a safe Rust interface for the PSA operation you added.
+
+## Frontend (Parsec)
+
+### parsec-book
+
+Create the specification page for the Parsec Book. Include operation details such as parameters,
```
hug-dev, line 37: I know this is annoying but could you add links of the different pages you are referring to 😬 (or examples of those)? So that it is easier for readers to just click on the page instead of searching themselves.

sbailey-arm: Yes no problem! This is still in draft mode on purpose!

src/contributing/adding_new_operation_how_to.md
sbailey-arm: Isn't `parsec-interface-rs` used to go between Protobuf and Parsec? (so it is always used?)

sbailey-arm: > We should also maybe put a note that the parsec-interface-rs and client parts are only for the Rust client but should be similar for other clients.

I thought about that, the reason I put `parsec-client-rust` as part of the procedure (i.e. not marked as optional) is the fact that Parsec's e2e tests require it, so it has to be done in order for Parsec to pass CI.

src/contributing/adding_new_operation_how_to.md
sbailey-arm: Could work, but we'd want to have a note saying what order it should be done in, right? The current order is the way I'd do it, now that I know how it all fits together and works.

src/contributing/adding_new_operation_how_to.md
hug-dev: Yeah both of those points are correct and you are right 😄

src/contributing/adding_new_operation_how_to.md
hug-dev: The order is the same to what you did, just the `psa-crypto*` crates grouped with `parsec` when doing the `Parsec Service` as I thought it would make sense to group those things together.

My main point was to give an idea about how section names would look like if you remove the `Frontend`/`Backend` names :)

sbailey-arm:
I've added a few (dozen...!) links to the page. I'm wondering about the ones that link straight to a line in code though. I preferred linking to a file, rather than a line, for the more 'latest version critical' bits (e.g. when picking the next opcode, you always want to be looking at the latest list of opcodes in use).

PR #47, opened 2020-07-10
Add a notion of priority in ListProviders

hug-dev:
The ListProviders now returns providers in order of priority.
See parallaxsecond/parsec-se-driver#7 for details.

### DIFF
--- a/src/parsec_client/operations/list_providers.md
+++ b/src/parsec_client/operations/list_providers.md
@@ -1,6 +1,6 @@
 # ListProviders
 
-Gets a list of available Parsec providers to be used by clients. Opcode: 8 (`0x0008`)
+Gets a prioritised list of available Parsec providers to be used by clients. Opcode: 8 (`0x0008`)
 
 ## Parameters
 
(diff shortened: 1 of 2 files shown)


src/parsec_client/operations/list_providers.md
```diff
@@ -1,6 +1,6 @@
 # ListProviders
 
-Gets a list of available Parsec providers to be used by clients. Opcode: 8 (`0x0008`)
+Gets a priority list of available Parsec providers to be used by clients. Opcode: 8 (`0x0008`)
```
ionut-arm, line 3: "prioritised"? 😅

src/parsec_client/operations/list_providers.md
```diff
@@ -36,6 +36,10 @@ The version triplet returned by this operation (`version_maj`, `version_min` and
 the implementation version of the specific Parsec provider. For the Core Provider, this version is
 the implementation version of the whole Parsec service.
 
+The `providers` vector returned is in order of provider priority: the highest priority providers
+come first. The core provider will always come last. The provider at position zero, if not the core
+provider, can be treated as default provider by the client.
```
ionut-arm, line 41: Do you think it's needed to mention here that the client needs to check supported opcodes too? "can be treated as default" sounds like "you don't need to check anything else". I guess for now it doesn't matter, it possibly will in the future!

hug-dev: Sure, will add a nuance.

PR #50, opened 2020-07-22
Update the installation guide

hug-dev:
* move the commands needing `sudo` at the beginning so that they are done by the admin
* inform that rust might need to be installed in the `parsec` user
* add an example command to add an user to the `parsec-clients` group
* inform that log-out/log-in might be needed
* inform that priviliges might need to be set for the `parsec` user

### DIFF
--- a/src/parsec_service/install_parsec_linux.md
+++ b/src/parsec_service/install_parsec_linux.md
@@ -15,28 +15,6 @@ can be a space or comma-separated subset of: `mbed-crypto-provider`, `pkcs11-pro
 `tpm-provider`. Choose the providers you want to install depending on what is available on the
 platform.
 
-Create and log in to a new user named `parsec`.
-
-```
-sudo useradd -m parsec
-sudo passwd parsec
-su --login parsec
-```
-
-In its home directory, pull and install Parsec as a daemon.
-
-```
-git clone https://github.com/parallaxsecond/parsec.git
-cargo install --features $DESIRED_FEATURES --path parsec
-```
-
-Copy and adapt the [configuration](configuration.md) you want to use. For a secure deployment, make
-sure to activate the `log_error_details` option and to use a `trace` log level.
-
-```
-cp parsec/config.toml config.toml
-```
-
 Create the Parsec socket directory.
 
 ```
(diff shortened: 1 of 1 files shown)


src/parsec_service/install_parsec_linux.md
```diff
@@ -15,48 +15,62 @@ can be a space or comma-separated subset of: `mbed-crypto-provider`, `pkcs11-pro
...
+that specific user.
 
 ```
-sudo chmod 755 /tmp/parsec
+git clone https://github.com/parallaxsecond/parsec.git
+cargo install --features $DESIRED_FEATURES --path parsec
+```
+
+Copy and adapt the [configuration](configuration.md) you want to use. For a secure deployment, make
+sure to activate the `log_error_details` option and to use a `trace` log level.
```
ionut-arm, line 70: It's probably not a good idea to change this right now, but I still don't think either `log_error_details` or `trace` help with security. The log flag would print more details of errors - potentially helpful if a problem occurs, but doesn't add anything to security (actually, as written in the config example file, it could lead to some data leakage). `trace` is a (very) weak protection against repudiation - while it might have some value, the sheer amount of data it would is probably not warranted in a production environment.

Again, not for now, we can refactor at some later point.

src/parsec_service/install_parsec_linux.md
```diff
@@ -15,48 +15,62 @@ can be a space or comma-separated subset of: `mbed-crypto-provider`, `pkcs11-pro
...
-sure to activate the `log_error_details` option and to use a `trace` log level.
+For example, adding the imaginary `parsec-client-1` user to the `parsec-clients` group:
 
 ```
-cp parsec/config.toml config.toml
+sudo usermod -a -G parsec-clients parsec-client-1
 ```
 
-Create the Parsec socket directory.
+Users just added to that group might need to log-out and log-in again to make sure the change apply.
```
paulhowardarm, line 40: I think we need to shout a bit louder about this login/logout cycle, because that would be easy to miss by someone who is following this recipe and mainly just looking for the commands that they need to type. Blurb in between the commands would be easy to glance over without really thinking about it. Perhaps flag it with a bold "Important", or better still add a paragraph right at the end (after referring the e2e_tests) with a cautionary note along the lines of: if you see PermissionDenied errors from the client, double check that the group change has taken effect, which you can check by calling the `groups` command with no arguments.

hug-dev:
> Apart from the actual permissions on the parsec user, can everything else be automated (maybe with passing a list of users as command arguments)?

I think so! The list of users to be in the `parsec-clients` group seems to be the only input.

Also the following commands:

```
git clone https://github.com/parallaxsecond/parsec.git
cargo install --features $DESIRED_FEATURES --path parsec
```
could be replaced with:
```
cargo install parsec-service --features $DESIRED_FEATURES
```
to install the latest version published on crates.io

src/parsec_service/install_parsec_linux.md
hug-dev: Agree that logging might not be the best mitigation against repudiation and would be ok to remove it if we find another mitigation to the related threats.

src/parsec_service/install_parsec_linux.md
hug-dev: Will change!

src/parsec_service/install_parsec_linux.md
ionut-arm: Unfortunately I don't think there's much we can do without mutual authentication 😞

src/parsec_service/install_parsec_linux.md
ionut-arm: > with a cautionary note along the lines of: if you see PermissionDenied errors from the client, double check that the group change has taken effect

On as similar note, maybe we should also create a troubleshooting page - "if you see _this_ error, do _this_" or "you haven't done _that_"

PR #52, opened 2020-07-27
Update parsec-book with documentation regarding `psa_generate_random`

joechrisellis:
Documentation updates for the `psa_generate_random` operation.


joechrisellis:
Note that the dead link message is because this is dependent on [this patch](https://github.com/parallaxsecond/parsec-operations/pull/15).

PR #53, opened 2020-08-05
Added psa_copy_key

sbailey-arm:


sbailey-arm:
@ionut-arm should we close this until copy is properly spec'd for Parsec?

PR #57, opened 2020-08-07
Add documentation for Unix domain socket authenticator

joechrisellis:


ionut-arm:
Changes in the API overview page are also needed, in the [Authentication and Sessions](https://parallaxsecond.github.io/parsec-book/parsec_client/api_overview.html#authentication-and-sessions) section.

src/parsec_service/authenticators.md
```diff
@@ -5,6 +5,23 @@
...
+context, 'peer credentials' refers to metadata about the connection between client and server that
+contains the effective user ID (UID) and group ID (GID) of the connecting process.
+
+To use this authenticator, the application must self-declare its UID in the authentication field of
+the request as a UTF-8 string. This authenticator will then verify that the UID sourced from the
+peer credentials matches the one self-declared in the request. If they match up, authentication is
+successful and the application identity is set to the UID.
+
+Currently, the peer credentials authenticator only supports authentication via the peer credentials
+sourced from a Unix domain socket. However, the authenticator will likely support different
```
paulhowardarm, line 22: Not sure about the phrasing here. It seems like the reverse of what we intend. Peer credential specifically requires domain socket to be the transport, because we rely on the ability of UDS to be able to query the credential in a trusted manner. I think it's misleading to say that peer credential would support other transports in the future, unless we're willing to significantly generalise our definition of what a peer credential is (in which case I don't think we can specifically talk about Unix UIDs or GIDs).

It would be more accurate to say the reverse: that a UDS transport could be used with other auth mechanisms and not only peer credential.

src/parsec_client/api_overview.md
```diff
@@ -177,6 +187,10 @@ permitted numerical values for this field are given as follows:-
 - A value of 2 (`0x02`) indicates authentication tokens. The service will expect the
    **authentication** field to contain a JWT token. Tokens must be signed with the private key of
    the identity provider and their validity period must cover the moment when the check is done.
+- A value of 3 (`0x03`) indicates peer credentials authentication. The service will expect the
+   **authentication** field to contain the stringified UID of the connecting process. The Parsec
```
paulhowardarm, line 191: "stringified uid" feels a bit vague here considering this section is supposed to be the specification of how a client process should populate the auth field. We don't mention that it has to be UTF8 (although that is mentioned somewhere further down, but I think it needs to be here). We also probably want to be very explicit that we expect a UID and **not** a user name. Also, since UIDs are numerical, we probably want to specify what number format we expect. Decimal? Hex? Will the service parse conventional notations such as the` 0x` prefix for hex numbers? This is the level of precision that I think we need here.

src/parsec_service/authenticators.md
ionut-arm: That might've been my fault, as I suggested expanding what we assume - the definition of "peer credential" would indeed depend on the transport mechanism, but I don't see an issue in assuming that other IPC mechanisms (e.g. network, d-bus) could be covered with this kind of authentication if we point out that only UDS are supported for now.

joechrisellis: Makes sense -- removed mentions of other transports for this section.

src/parsec_client/api_overview.md
joechrisellis: Good point -- I've changed this to:

```
- A value of 3 (`0x03`) indicates peer credentials authentication. The service will expect the                                                                                                                                                                                                                                                                                                                                           
  **authentication** field to contain the non-padded decimal UID (**not** username) of the
  connecting process as a UTF-8 string (e.g. `123`). The Parsec service will verify that this
  self-declared UID is consistent with the UID from the peer credentials.
```

Please let me know if there's any further suggestions!

src/parsec_service/diagrams/interfaces_and_dataflow.drawio
```diff
@@ -1 +1 @@
...
...oVuarmjpdNZVbt18EEv6fLogwnHXKg6y7rVUeny+LLbp0LbfyG3FBjDLWuXXxjxqQ+Qu9vLv74ATO8oJsETpQd3xbREi9enF2xSC7ppExrcYIGRantmGCRZa/ERwNXQpjUR5zaI9rsoBztlkGbZewF6cTgZ+tKftR09lmpj23dEqzbPjJ2ioBz25fI2/jn6e/3aaSkCBnn3ujrME/joLDy4QdKEjq0VtR+/60R3Ml10/qLG3gtApq0GjTTDBM2qiKQfWkjzvbdSmk3Vv+cEtevHUi9wljZBYXVfGvSE1AD6W5lXmw9y7y0qVZPSzvSSnk1O4j5a0CJJxKJ05pAJRVM6rKus4VcwZoBvlBYo5bbnQEMSeuYpBIMddZp+fTclpKlapoGr1KRRlcO8iyme1I6dwFycJOSplUrrB7SJeXxsV6lS9JGDlLwcU2SJ3O+YvvBW4mfNrqKOdS8P1jW2HlHvZEn0WpVAYTL8GFWyhZxJYve5eEyYfDbbbgONm5laYwrx2/yJMGATt9ySrytLdQm3wIlBGI13pfPCjw7LVx3Y57yjcN5Tzf/awohpHKtNYFSEJlTy9wdXMQYKJ6mxgNwltGAp5KLQBe0LbpyCan9LbNUZLJFibiwSVgVewUwfU+U3IuCFvANBgl9sjYIpLz3LBVWR/Vb/HuaSOMygMHgGPiqxx8iVjizceXgFybKeIrj/RmQd/4gvBlUdke5a4kVYBpPM1Loy1/cRAD5wmxYtnvBaGLk1VGtolusfM1ITsk8+ElK5noNkyZqyxRQG2hdVaxNaoKHHHatZQcARL5KTBd/bDIWkfX8VBfdSPzWl3OJK7T6nImkaupuJyZmlu6Whaz6x1dg3RZsZGja4S4IzicMdUXYbTGF5BAm2AH0ypcP25jcdIKXDHhsvJKAUJLhXRkVdlVRayqOry5yJZlq5my1lSWKc1FHPUzijdhEKOsx2g7T7ZRFmWdR+h8J6oZFSGFWGXlRUgBh+XAljZdmc1hp8OY/t4Jlltsw1RD6pW9AbAKHMTb9Ya/t+d5IkClrJtZ6UQ+JgCaowwH6p9ZrbMq+32yy+IFMp/eL2+m2gZvIcPY3JeV0Ufm/IbhfxE8V0tulmFl1jzhujNkclxWDcxyD9XL47NC8VkjPqMWtwLoHzrGh1GY0n1nCmB6rD5gYz+94h8=</diagram></mxfile>
```
hug-dev, line 1: Thanks for updating the diagram as well! Could you also please keep the white background? I believe that in the png you exported from draw.io the background is transparent. Just so that it is easier to read on our book 😃!

hug-dev (commented):
Thanks, that looks good to me 😃 One comment about the PNG

src/parsec_client/api_overview.md
hug-dev: I think to answer to your original question about the format, we need to be clear about the scope of the authentictor. Is it a generic peer credentials authenticator which could be used with other operating systems/listener types or a **Unix** peer credential authenticator?
If the authenticator that we are describing here is only the later then I think we should:
* be more precise where we mention "peer credentials" and always say "Unix peer credentials" and make sure that when we say "UID" we mean the specific Unix User identifier
* the Unix User Identifier is represented with [an integer type](https://en.wikipedia.org/wiki/User_identifier#Type). Most Unix OS use 32 bits for this, and Linux uses the `uid_t` type which is an unsigned 32-bits number. For that reason, I think it would be better to expect the authentication payload to be the representation of a `u32` value, 4-bytes padded with zeroes. For example the UID `4` would be `[0x00, 0x00, 0x00, 0x04]` (or reverse depending on our endianess). I think this is more efficient and less error-prone than having to parse/format a UTF-8 string.

src/parsec_service/diagrams/interfaces_and_dataflow.drawio
joechrisellis: I just checked the other `.drawio` files in the repository and they all seem to have transparent backgrounds as well? Am I missing something?

src/parsec_client/api_overview.md
joechrisellis: > I think to answer to your original question about the format, we need to be clear about the scope of the authentictor. Is it a generic peer credentials authenticator which could be used with other operating systems/listener types or a Unix peer credential authenticator?

AFAIK 'peer credentials' is a Unix-exclusive thing, but I'm all for being more explicit. 😄 

> be more precise where we mention "peer credentials" and always say "Unix peer credentials" and make sure that when we say "UID" we mean the specific Unix User identifier

ACK, will do.

> the Unix User Identifier is represented with an integer type. Most Unix OS use 32 bits for this, and Linux uses the uid_t type which is an unsigned 32-bits number. For that reason, I think it would be better to expect the authentication payload to be the representation of a u32 value, 4-bytes padded with zeroes. For example the UID 4 would be [0x00, 0x00, 0x00, 0x04] (or reverse depending on our endianess). I think this is more efficient and less error-prone than having to parse/format a UTF-8 string.

Agreed, that's a lot nicer. Will do. 🙂

src/parsec_service/diagrams/interfaces_and_dataflow.drawio
joechrisellis: Actually, you're right -- the `.drawio` files _do_ have transparent background, but I had selected the 'export with transparent background' option.

hug-dev (approved):
Thanks for the changes, looks good to me!!

src/parsec_service/authenticators.md
```diff
@@ -5,6 +5,25 @@
...
+authentication field of the request as a zero-padded little-endian 32-bit unsigned integer. This
+authenticator will then verify that the UID sourced from the peer credentials matches the one
+self-declared in the request. If they match up, authentication is successful and the application
+identity is set to the UID.
+
+Note that a Unix domain socket transport is not limited to the Unix peer credentials authenticator;
+this transport can be used with a different authenticator if required.
+
+The GID component of the Unix peer credentials is currently unused by the peer credentials
+authenticator.
```
ionut-arm, line 27: Gonna guess the PID isn't used either.

ionut-arm (approved):
👌

PR #60, opened 2020-08-18
Add documentation for ListAuthenticators operation

joechrisellis:


src/parsec_client/operations/list_authenticators.md
```diff
@@ -0,0 +1,43 @@
+# ListAuthenticators
+
+Gets a list of available Parsec authenticators available for use. Opcode: 14 (`0x0014`)
```
hug-dev, line 3: Seems like you repeat "available" twice 😛 Also, I think the list returned by this operation is ordered? Last thing, it might be nice to indicate "available for use at the Listener endpoint" at the end to inform it is only for the listener endpoint used to perform the request.

src/parsec_client/operations/list_authenticators.md
```diff
@@ -0,0 +1,43 @@
...
+## Specific response status codes
+
+No specific response status codes returned.
+
+## Description
+
+The version triplet returned by this operation (`version_maj`, `version_min` and `version_rev`) is
+the implementation version of the specific Parsec authenticator.
+
+The `authenticators` vector returned is in no specific order. The core authenticator will always
```
hug-dev, line 36: I think this operation should be ordered, the first authenticator returned being considered as the primary one.

hug-dev (commented):
Thanks! A few comments.

src/parsec_client/operations/list_authenticators.md
joechrisellis: Good spot on the double-use of 'available' -- thanks for that!

And yes, this operation is ordered, I just wasn't sure if we had a specific ordering in mind. I agree with your other comment, though -- we should order the authenticators by priority (i.e. the primary authenticator comes first).

Will amend!

src/parsec_client/operations/list_authenticators.md
joechrisellis: ACK, will amend.

hug-dev (approved):
Looks good!

PR #70, opened 2020-10-06
Add the ListKeys operation

hug-dev:

### DIFF
--- a/src/parsec_client/operations/list_keys.md
+++ b/src/parsec_client/operations/list_keys.md
@@ -1,0 +1,37 @@
+# ListKeys
+
+Lists all keys belonging to the application. Opcode: 26 (`0x001A`)
+
+## Parameters
+
+No parameters are needed for this operation.
+
+## Results
+
+| Name   | Type                                 | Description             |
+|--------|--------------------------------------|-------------------------|
+| `keys` | Vector of [`KeyInfo`](#keyinfo-type) | List of key information |
+
+### KeyInfo type
+
+A `KeyInfo` type contains the following members:
+
+| Name          | Type                                                        | Description                        |
+|---------------|-------------------------------------------------------------|------------------------------------|
+| `provider_id` | Unsigned integer                                            | ID of the provider holding the key |
+| `name`        | String                                                      | Name of the key                    |
+| `attributes`  | [`KeyAttributes`](psa_key_attributes.md#keyattributes-type) | Attributes of the key              |
+
+## Specific response status codes
+
+No specific response status codes returned.
+
+## Description
+
+This operation lists all the keys that an application created in all providers.
+
+## Contract
+
+No contract yet.
+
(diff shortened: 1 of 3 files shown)


src/parsec_client/operations/list_keys.md
```diff
@@ -0,0 +1,37 @@
+# ListKeys
+
+Lists all keys belonging to the application in all providers. Opcode: 26 (`0x001A`)
```
ionut-arm, line 3: Would it make sense to have this work per-provider, and if it's sent to the `CoreProvider` it returns all keys from all crypto providers?

hug-dev: That would be a nice addition I think. Would we still keep the same `KeyInfo` structure? Because the `id` field (I have to rename it `provider_id`) would be the same then

ionut-arm: Yeah, why not, clients can just ignore that data if they don't want it.

hug-dev: Cool, done!

src/parsec_client/operations/list_keys.md
```diff
@@ -0,0 +1,37 @@
...
+|--------|--------------------------------------|-------------------------|
+| `keys` | Vector of [`KeyInfo`](#keyinfo-type) | List of key information |
+
+### KeyInfo type
+
+A `KeyInfo` type contains the following members:
+
+| Name          | Type                                                        | Description                        |
+|---------------|-------------------------------------------------------------|------------------------------------|
+| `provider_id` | Unsigned integer                                            | ID of the Provider holding the key |
```
joechrisellis, line 21:
```suggestion
| `provider_id` | Unsigned integer                                            | ID of the provider holding the key |
```

src/parsec_client/operations/list_keys.md
```diff
@@ -0,0 +1,37 @@
...
+| `name`        | String                                                      | Name of the key                    |
+| `attributes`  | [`KeyAttributes`](psa_key_attributes.md#keyattributes-type) | Attributes of the key              |
+
+## Specific response status codes
+
+No specific response status codes returned.
+
+## Description
+
+This operation lists all the keys that an application created in all providers.
```
joechrisellis, line 31: Does this want to be changed re @ionut-arm's suggestion?

joechrisellis (commented):
Few minor comments from me!

src/parsec_client/operations/list_keys.md
hug-dev: We decided internally that since `ListKeys` is a core operation, it should only target the Core provider 😄

src/parsec_client/operations/list_keys.md
hug-dev: Done! Thanks.

ionut-arm (approved):

PR #74, opened 2020-10-15
Update the book after security changes

### LINKED ISSUE #73: Update the book after enablement of Unix Peer Crendentials Authenticator
Check that the content is still correct and up to date after the Unix Peer Credentials Authenticator has been enabled throughout the stack and especially that the security checks in the client have been removed.

hug-dev:
Since the socket folder has been moved to /run/parsec, the clients no
longer have to check for a parsec owner. Update the related book pages
to reflect that.

Also update the TM after the Unix Peer Credentials auth was added.

Close #73


src/parsec_security/secure_deployment.md
```diff
@@ -2,25 +2,31 @@
...
+- In a deployment using **Direct authentication with mutually trusted clients**, the socket folder
+   must be group owned by the `parsec-clients` group which has only read and execute permission on
+   it (O-10). Everyone else must have no permission on the folder. The `parsec-clients` group is
+   composed of mutually trusted clients only. This group must be continuously maintained to contain
+   the mutually trusted clients.
+- In a deployment using **Unix Peer Credentials authentication**, everyone else can have read and
+   execute permission on the socket folder.
+- In a deployment using **Unix Peer Credentials authentication**, when a Unix account is deleted,
+   all of its Parsec keys must also be deleted (O-11).
+- The Parsec configuration must not be reloaded with a different authenticator is Parsec is
```
ionut-arm, line 28:
```suggestion
- The Parsec configuration must not be reloaded with a different authenticator if Parsec is
```

src/parsec_security/secure_deployment.md
```diff
@@ -2,25 +2,31 @@
...
+   currently storing keys (O-12).
```
ionut-arm, line 29: Maybe add what should be done instead? Like, "remove all contents of `/var/lib/parsec/mappings/`"

src/parsec_service/build_run.md
```diff
@@ -3,51 +3,35 @@
...
+Parsec from the current directory, have the key information mappings in `./mappings` and the socket
+at `/tmp/parsec.sock`. The test configuration will make those choices.
 
 Having cloned the Parsec repository, to build and run from source using the Mbed Crypto provider and
 the test configuration:
 
 ```````
-RUST_LOG=info cargo run --features mbed-crypto-provider -- -c e2e_tests/provider_cfg/mbed-crypto/config.toml
+cargo build --features mbed-crypto-provider
+RUST_LOG=info ./target/debug/parsec -c e2e_tests/provider_cfg/mbed-crypto/config.toml
```
ionut-arm, line 27: Why separate it like this?

src/parsec_security/parsec_threat_model/threat_model.md
```diff
@@ -147,27 +147,27 @@ mitigation status and the assets impacted. A threat can be unmitigated (U), miti
...
+| S | (GNRC) An attacker uses brute-force on the AUTH field of the request until he does not have an AuthenticationError response status to then deduce a valid authentication token.             | M-2                              | AS1      |
+| S | (DSPC) An attacker uses another client's Unix account. He can then have access (reference) to any of the impersonated client's keys and execute any operation in their name.                | U-5                              | AS3      |
+| S | (DSPC) An attacker uses a Unix account with the same UID as a client. He can then have access (reference) to any of the impersonated client's keys and execute any operation in their name. | O-11, ASUM-3                     | AS3      |
+| S | An attacker uses a weakness or a bug of one authentication method as soon as the authentication method used by Parsec is changed.                                                           | O-12, U-6                        | AS3      |
```
ionut-arm, line 162: I'm not sure I understand this one - more precisely, I don't get what `as soon as the authentication method used by Parsec is changed` means.

src/parsec_security/secure_deployment.md
```diff
@@ -2,25 +2,31 @@
...
+   `parsec` user (O-9).
+- In a deployment using **Direct authentication with mutually trusted clients**, the socket folder
+   must be group owned by the `parsec-clients` group which has only read and execute permission on
+   it (O-10). Everyone else must have no permission on the folder. The `parsec-clients` group is
+   composed of mutually trusted clients only. This group must be continuously maintained to contain
+   the mutually trusted clients.
+- In a deployment using **Unix Peer Credentials authentication**, everyone else can have read and
+   execute permission on the socket folder.
+- In a deployment using **Unix Peer Credentials authentication**, when a Unix account is deleted,
+   all of its Parsec keys must also be deleted (O-11).
```
ionut-arm, line 27: I think you'll need to mention that currently this must be done using the `parsec-tool` to list and manually remove all the keys. We could add a command in the tool for this. But in the future there might be an operation available for clients/admins to remove these keys.

src/parsec_security/parsec_threat_model/threat_model.md
```diff
@@ -322,20 +322,22 @@ cleared.
...
+| 4  | While Parsec is authenticated on the device, anyone can brute force a key ID to execute operations with the key. Unmitigated for the PKCS 11 Provider: all sessions share the login state; if one session logs in then all other opened sessions will also be logged in. Other sessions only need a valid key ID to execute operations with private keys. Mitigated for the TPM Provider: each key is protected by a long, random authentication value, generated by the TPM | All the client's PKCS 11 keys can be used with PKCS 11 operations.                                                         |
+| 5  | (DSPC) Parsec does not know about the security quality of a client's Unix password.                                                                                                                                                                                                                                                                                                                                                                                          | All the client's keys can be used, and exported if allowed by the key's policy.                                            |
```
ionut-arm, line 331: Can you make the "conditional" stuff more consistent - i.e. the authenticator "code" should be italic and it should either always be at the beginning or at the end. At the beginning might be better, so you know as soon as you start reading.

ionut-arm (commented):
Cheers 🎉 lots of nice changes

src/parsec_security/secure_deployment.md
ionut-arm: Ergh, maybe telling people to remove their mappings ain't the best idea. But the point stands - why not or what should they do instead? Maybe use a different folder?

src/parsec_security/parsec_threat_model/threat_model.md
hug-dev: That is in the context of Parsec supporting only one authenticator. If you change the authenticator to a weaker one containing a security bug, then all of your keys are as secure (for spoofing) as this authenticator is. And that happens as soon as the authenticator is changed.

src/parsec_security/secure_deployment.md
hug-dev: I will add a sentence to explain how it could be done, either by the `parsec-tool` or by the Rust client as well.

src/parsec_security/secure_deployment.md
hug-dev: Yes good point, I will add that a new folder can be chosen, or all keys should be deleted first.

src/parsec_service/build_run.md
hug-dev: It seemed to me more idiomatic for the build and run steps to be separate. It also makes the lines shorter on the book, so you don't have to scroll horizontally on certain resolution.

src/parsec_security/secure_deployment.md
ionut-arm: I'm a bit nervous about the "delete all keys" option - should deffo have a "WARNING" in front. ⚠️

hug-dev:
Adressed your comments and also updated the list of existing tests as the links were wrong because of recent changes.

There are some link errors but they are all due to "Too Many Requests" which I think we can ignore.

ionut-arm (approved):
📨 Let's get this IN!

PR #75, opened 2020-11-06
Add documentation related to the JWT-SVID auth

hug-dev:
Updates the authenticator sections and the thread model.


hug-dev:
`OM-14` can be deleted when we decide to do local validation of JWT-SVID tokens by fetching the trust bundle (the SPIFFE public keys) through the Bundle Endpoint which is authenticated.
It was agreed that we would do that in a second phase, and that I open a GitHub issue for that problem when the PRs are merged.

ionut-arm:
Hmm, I have one question related to the diagram - do we not consider the Spiffe implementation to be trusted like the generic "Identity Provider"? Is it not possible to combine the two within the diagram and threat model, and just mark the ones that are Spiffe or "Generic" related with their respective codes?

hug-dev:
> do we not consider the Spiffe implementation to be trusted like the generic "Identity Provider"? Is it not possible to combine the two within the diagram and threat model, and just mark the ones that are Spiffe or "Generic" related with their respective codes?

Yes good question, I thought about that as well. I think I preferred put them apart because ideally we should not make any assumption about the SPIFFE implementation, we should not have to trust it.

This is kind of a "temporary" solution where the admin has to trust the SPIFFE Workload API Endpoint but in the future it might be possible to not trust it completely if the validation happens locally.

ionut-arm:
> we should not have to trust it.

I mean, we *have* to trust anything that's an identity provider, otherwise the model doesn't make sense. For example, if the Spiffe implementation is malicious, there is literally nothing that you can do (not as things stand). 

That being said, maybe we need to create a new yellow "trusted" box with "OS" inside. We deffo need that to be trusted.

hug-dev:
> For example, if the Spiffe implementation is malicious, there is literally nothing that you can do (not as things stand) to prevent abuse by some clients.

Yes, in the current implementation I agree. However, the SPIFFE public keys can be fetched using another special endpoint which is authenticated and backed up by CA (as far as I know). So you still have to trust something but it is something commonly accepted

ionut-arm:
Ah, that's fair!

src/parsec_security/parsec_threat_model/threat_model.md
```diff
@@ -320,6 +326,35 @@ cleared.
...
+
+This attacker uses the SPIFFE Workload API endpoint created by the SPIFFE implementation to
+interfere with the validation of the JWT-SVID token sent by the client. It can also create a spoofed
+endpoint, mimicking the SPIFFE implementation's one.
+
+*Only applicable for deployment (JWTS).*
+
+|   | Description                                                                                                                | Mitigation | Assets                  |
+|---|----------------------------------------------------------------------------------------------------------------------------|------------|-------------------------|
+| S | An attacker impersonates the SPIFFE implementation: Parsec sends the authentication data of clients to malicious entity.   | O-14       | AS1                     |
```
ionut-arm, line 352: Isn't what you describe here more for A12? Would've said that spoofing for the response means the attacker sending their own response to the service. 

Tbh, I'm not sure for communication channels like this or between client and service if having separate attackers for each direction makes that much sense 😕

ionut-arm (commented):
Nice stuff, just some nitpicky comments below 🤓

src/parsec_security/parsec_threat_model/threat_model.md
hug-dev: hmm for me A12 is when the attacker either mimick the Parsec service or spy on the request and A13 is when the attacker mimicks the SPIFFE implementation or spy on the response.

Under this definition I did not add a lot of threats for A12 because then it would be doing the SPIFFE implementation threat model.

I agree though that it might not make sense to split this into two and is a bit messy... Maybe in the future we want to merge our attackers. For this one I can either merge now or leave it like that?

ionut-arm: Yeah, I think it makes sense to merge them in the future, I can't understand why we created attackers that are mostly in other project's threat model.

src/parsec_security/parsec_threat_model/threat_model.md
```diff
@@ -330,20 +365,22 @@ cleared.
...
 | 6  | Parsec does not support using multiple authenticators concurrently ([open issue](https://github.com/parallaxsecond/parsec/issues/272)).                                                                                                                                                                                                                                                                                                                                      | If there is weakness in one authenticator, all the client's keys can be used, and exported if allowed by the key's policy. |
+| 7  | *(JWTS)* The validation steps should parse the JWT-SVID token and make sure the claims of the client's token are the same than the ones returned by the validation operation from the Workload API.                                                                                                                                                                                                                                                                          | All the client's keys can be used, and exported if allowed by the key's policy.                                            |
```
ionut-arm, line 368: > validation steps should parse the JWT-SVID token and make sure the claims of the client's token are the same than the ones returned by the validation operation from the Workload API.

Isn't this something that we should be doing in the service?

ionut-arm (approved):
Thanks, looks good overall - I still have one question, though it's more related to the service PR.

src/parsec_security/parsec_threat_model/threat_model.md
hug-dev: I believe this validation step should be common to all JWT-SVID token so I think it could be in `rust-spiffe`. This is basically to check that the SPIFFE ID that we are validating is the same as the one that is returned by the validation operation.

ionut-arm: Hmm, even though we're working on that repo too, I don't think we need to include that here. I'm sure there are lots of cases like this in other libraries we use, for other data flows

hug-dev: Ok, yeah, will remove

hug-dev:
Removed U-7 and added a link to parallaxsecond/parsec#289 as said [here](https://github.com/parallaxsecond/parsec-book/pull/75#issuecomment-723195651).

hug-dev:
Merging as the only errors are "Too Many Requests" but the links are valid.

PR #77, opened 2021-01-05
Add upstreaming guide

ionut-arm:
This commit adds a guide for contributors who look to upstream one of
the projects under the Parsec umbrella.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -60,5 +60,6 @@
 - [Contributing](contributing/README.md)
    - [Pull request checklist](contributing/pr_checklist.md)
    - [How to add operations to Parsec](contributing/adding_new_operation_how_to.md)
+   - [Package management and versioning guide](contributing/package_management.md)
 
 *Copyright 2019 Contributors to the Parsec project.*
--- a/src/contributing/README.md
+++ b/src/contributing/README.md
@@ -27,4 +27,7 @@ Signed-off-by: Full Name <email address>
 If you'd like to get a feel for what we're looking for in contributions, we also have a checklist
 for PR reviewers [here](pr_checklist.md).
 
+When planning to publish code to a package manager, make sure to read our [package management and
+versioning guide](package_management.md).
+
 *Copyright 2019 Contributors to the Parsec project.*
--- a/src/contributing/package_management.md
+++ b/src/contributing/package_management.md
@@ -1,0 +1,103 @@
+# Package management and versioning guide
+
+This page presents the details around packaging management and versioning for Parsec-related
+repositories. Currently our repos are mainly Rust crates, so this guide is accordingly geared
+towards them - when other guidelines become relevant, e.g. specific to Go libraries, new sections
+should be added below.
+
+The following points are relevant to all software we publish to package management repositories:
+
+- For published versions of a project a tag carrying the version number should be upstreamed to
+   GitHub, e.g. if version 2.3.1 was published, a tag named "2.3.1" must be added to the commit in
+   GitHub.
+- If multiple projects that end up in a package manager reside in the same repo, the tags should
+   reflect both the project and the version number, e.g. if version 2.3.1 of project X was
+   published, a tag named "X-2.3.1" or similar must be used.
+- Projects should follow [Semantic Versioning](https://semver.org/) rules.
+- Following the release of a new version of publicly-facing projects, update the `CHANGELOG.md`
+   file. For some projects it's also advised to create a GitHub Release from the tag mentioned
+   above.
+
+## Rust crates
+
+The process around publishing Rust crates is tightly linked with crates.io. In order to publish or
+have control over crates, you must own an account on the website. Before you decide whether you want
+to publish a crate, check out the [Dependency management](#dependency-management) section below.
+
+- Follow the instructions [here](https://doc.rust-lang.org/cargo/reference/publishing.html) to
+   create your account and to set your workspace up.
+- If you'd like to publish a new version of an existing crate using the ownership granted to the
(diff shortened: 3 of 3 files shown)


ionut-arm:
😠 crates.io links don't work, forgot about that. 

Oh, and I forgot to mention the problems with docs.rs

src/SUMMARY.md
```diff
@@ -60,5 +60,6 @@
 - [Contributing](contributing/README.md)
    - [Pull request checklist](contributing/pr_checklist.md)
    - [How to add operations to Parsec](contributing/adding_new_operation_how_to.md)
+   - [Upstreaming guide](contributing/upstreaming.md)
```
hug-dev, line 63: Is "upstreaming" the correct word here? I would think more of "publishing" or "releasing"? "upstreaming" for me refers to pushing anything to the public, ie making pull-requests.

src/contributing/upstreaming.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Rust crates
+
+The process around publishing Rust crates is tightly linked with crates.io. In order to publish or
+have control over crates, you must own an account on the website.
+
+- From the main page of crates.io you can log in using your GitHub account.
+- Go to your account settings page and generate a new API token - this will be used on the command
+   line for publishing the crates.
+- Using the token generated above, log in on the command line: `cargo login <token>`.
```
hug-dev, line 29: Maybe these steps can be replaced linking the instructions from [this page](https://doc.rust-lang.org/cargo/reference/publishing.html) (section Before you first publish)?

hug-dev (commented):
A few nitpicks but otherwise it looks good to me!

src/SUMMARY.md
ionut-arm: I can change that - to me "upstreaming" is more general, moving something to somewhere higher in the consumption/distribution chain: local copy -> own fork -> original repo -> package manager. When other people use our code, they pull it from crates.io, not from GitHub (generally)

ionut-arm: (actually, it seems "publishing" is derived from the Latin word for "making public", I guess they're all kinda saying the same thing in different words)

hug-dev: It's ok then!

src/contributing/upstreaming.md
ionut-arm: Dun? I didn't link that section, I think reading the whole page is a good idea anyway

hug-dev (approved):
Looks good, thank you very much, very much indeed

src/SUMMARY.md
paulhowardarm: I think I agree with Hugues here - I think "upstreaming" is the wrong term, and this could be quite confusing to people who are just making a regular PR into master without any complex interface or dependency updates. I would call it a versioning guide, or a component versioning guide, or a versioning and dependency management guide.

paulhowardarm: Also think that we should refer to the pieces as "components" rather than "projects".

paulhowardarm (commented):
Fantastic to see this being added - I think it fills an important gap.

Per my other comment, I would agree with @hug-dev that I think "upstreaming" isn't the right term here. This is really about component versioning and dependency management. And on the same topic, I think what's missing is some additional high-level description of the overall strategy, particularly with respect to when it's okay to use git references versus when we need to publish a versioned crate. Perhaps we could do that with reference to some typical examples. For instance, it will be common for people to add to `parsec-interface-rs` and then need to consume this into `parsec` and `parsec-client`. Perhaps a bit of scene-setting about which crates are the "leaf" components, and an informal description of what the dependency tree looks like (perhaps with a link to other parts of the book that talk about the code structure in more detail).

ionut-arm:
I have updated the parts that were already there, but I've not added a new section because I don't think we agreed on a way to deal with transient dependencies - whether we always try to have all dependencies published, or if it's ok to stay in "limbo" between versions. I was thinking of using the minor version numbers for that, but it would prohibit us from making some change in an API, for example, and not committing to it right away (if we want some other change done in parallel used by dependents). Maybe we don't have to commit to either of those right now and just say "wheneve there's a need for a new vesion release, discuss it in the public channels first".

src/SUMMARY.md
```diff
@@ -60,5 +60,6 @@
 - [Contributing](contributing/README.md)
    - [Pull request checklist](contributing/pr_checklist.md)
    - [How to add operations to Parsec](contributing/adding_new_operation_how_to.md)
+   - [Upstreaming guide](contributing/package_management.md)
```
paulhowardarm, line 63: The link still reads as "Upstreaming guide" here - is that intentional?

src/contributing/README.md
```diff
@@ -27,4 +27,7 @@ Signed-off-by: Full Name <email address>
 If you'd like to get a feel for what we're looking for in contributions, we also have a checklist
 for PR reviewers [here](pr_checklist.md).
 
+When planning to upstream a project, make sure to read our [package management and versioning
```
paulhowardarm, line 30: Again we still have the "upstream" and "project" terms here, which I think we were moving away from.

ionut-arm: 🤦🏻‍♂️ yes, need to change those as well

paulhowardarm (approved):
Looks good to me! Thanks for contributing this.

ionut-arm:
I have added a subsection at the bottom about dependency management, let me know what you think about that as well - apologies for the delay in making that available

hug-dev (approved):
Thanks for the clarification of dependency tree

PR #85, opened 2021-02-03
Change content type and accept type protobuf value to 0x00

jn9e9:
Signed-off-by: John 9e9 <jn9e9e9@gmail.com>


paulhowardarm (approved):
Thanks for calling out the mistake. Doc fix looks good. I just want to figure out why the link checker is failing before we merge. There aren't any link changes in the PR, so it should be fine.

paulhowardarm:
The broken link is actually in a completely different part of the book, so we can fix that separately.

PR #93, opened 2021-02-23
Unmitigate repudiation threats

### LINKED ISSUE #84: Threat Model: review repudiation threats
The only way we currently have to fight againt repudiation threats is logging. That might be a bit foolish of us to think that's enough in most cases. It might be better to remove the requirements on providers to log each of their command and say that this threat in "unmitigated" until we have a better solution.

See parallaxsecond/rust-tss-esapi#86

hug-dev:
It is to better reflect the level of support we do through logging of
repudiation threats, and to be more honest.

Fix #84


src/parsec_security/parsec_threat_model/threat_model.md
```diff
@@ -359,29 +359,29 @@ endpoint, mimicking the SPIFFE implementation's one.
...
+| 3  | Listener implementations use an IPC mechanism respecting confidentiality and integrity of messages transmitted between the clients and the service (once the initial connection has been made).      | Unix Domain Socket Listener: the sockets used on the client and service side for the communication are represented by file descriptors that are only accessible by those processes.                                 |
+| 5  | Parsec is coded with safety in mind and is tested extensively.                                                                                                                                       | [Done](https://github.com/parallaxsecond/parsec-book/issues/35)                                                                                                                                                     |
```
ionut-arm, line 380: Hmm, I guess rebasing would confuse someone who already knew the threat model, but otherwise it would be ok to make `M-5` => `M-4`, `M-6` => `M-5` and so on

hug-dev: Yes I did not do it because of other ressources that might point to specific numbers :/ Also does not seem that it matters too much.

ionut-arm (approved):

Issue #90, opened 2021-02-11
General update of the book


hug-dev:
I have spotted a few things that need updating on the book, feel free to add more:

- [x] alignment of bullet points in the description of [keys](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_export_public_key.html)
- [ ] add documentation about the new providers
- [ ] the [dataflow diagram](https://parallaxsecond.github.io/parsec-book/parsec_service/interfaces_and_dataflow.html#data-flow-diagram) still refers to `KeyIDManager` instead of `KeyInfoManager`
- [ ] cross-compilation information
- [ ] code coverage information
- [ ] authenticators were added to features

hug-dev:
@ionut-arm Should we add the `KeyInfoManagerClient` in the diagrams?

ionut-arm:
I don't think so, it's more of a convenience layer that does some interfacing between KIM and providers

PR #95, opened 2021-03-02
Fix formatting and add clarifications from 1.0.1

hug-dev:
Which is the newest version of PSA Crypto.

Part of #90

Note that PSA Crypto 1.0.1 also added `PSA_ALG_SM3` and `PSA_KEY_TYPE_SM4` (full diff [here](https://armmbed.github.io/mbed-crypto/html/appendix/history.html)).
If we wanted to support those new alg/key type, we would have to modify our contract to add new variants in the `oneof` field. I am not sure if adding it would be a breaking change or not ([related](https://developers.google.com/protocol-buffers/docs/proto3#oneof)). It will definitely be a breaking change in the Parsec Rust Interface as those as strongly typed enums but it might be fine there? Maybe worth opening an issue for it.

### DIFF
--- a/src/parsec_client/operations/psa_export_key.md
+++ b/src/parsec_client/operations/psa_export_key.md
@@ -27,25 +27,27 @@ that is equivalent to key.
 
 For standard key types, the output format is as follows:
 
-- For symmetric keys, including MAC keys, the format is the raw bytes of the key.
+- For symmetric keys, including HMAC keys, the format is the raw bytes of the key.
 - For [`DES`](psa_key_attributes.md#des-type), the key data consists of 8 bytes. The parity bits
    must be correct.
 - For [`Triple-DES`](psa_key_attributes.md#des-type), the format is the concatenation of the two or
    three DES keys.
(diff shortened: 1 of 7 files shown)


src/parsec_client/operations/psa_export_key.md
```diff
@@ -27,25 +27,27 @@ that is equivalent to key.
 
 For standard key types, the output format is as follows:
 
-- For symmetric keys, including MAC keys, the format is the raw bytes of the key.
+- For symmetric keys, including HMAC keys, the format is the raw bytes of the key.
```
hug-dev, line 30: Note that the [spec differentiates](https://armmbed.github.io/mbed-crypto/html/api/keys/management.html#psa_export_key) between "HMAC keys that are longer than the underlying hash algorithm block size" and those that are equal or tinier.
Here that would make the choice that the format would always be the raw bytes of the key.

hug-dev:
Failed links are working correctly when testing manually

ionut-arm (approved):
Nice, I like the new text, thorough and clear.

PR #98, opened 2021-03-23
Add a Getting Started section

### LINKED ISSUE #89: Simplify Parsec onboarding
A lot of new members, interested by the Parsec idea, come and ask "Cool, so how do I get started?"

We can not expect people to conscientiously go through all of the book to get an idea of what Parsec is and how it is working. I propose that we write a new top-level sections (maybe in place of "Parsec for users") with the name "Getting started!" or something similar. This section would be the entrypoint we would send to anyone new to the project.

The goal would be to guide them through simple steps to have a Parsec service running and do some commands on the client side. It will help them having a global idea of what Parsec is and then we can point them to more advanced steps: running on real hardware, securely installing the service, advanced client libraries, customised configuration, etc...

I propose the following structure (I took example from the [SPIRE guide](https://spiffe.io/docs/latest/try/getting-started-linux-macos-x/)). Those should probably be split amoung different sub-pages for clarity.

# Getting Started!

## Overview

Explaining what we are going to do: running the parsec service + doing client commands

## Prerequisites

Should be as low as possible.

## Running the Parsec service

Multiple options there, depending on the providers:
* Mbed Crypto natively (the easiest)
* TPM provider in a Docker container
* PKCS11 provider in a Docker container

## Making Parsec requests

Multiple options as well:
* using the Parsec tool (signing something for ex.)
* performing end to end tests

## Next steps

* Check the Parsec clients (Rust, Go, C)
* Check how to install Parsec securely
* Try Parsec on real-HW providers
* Watch a Parsec conference, read the rest of the book!

We already have most of the information but it is scattered through our book.
We would delete the "How to build and run Parsec guide" after that.

Our "How to test Parsec" needs to be revised as well, and I am not sure if it makes sense. Testing Parsec locally is harder and harder because it depends of specific configurations.

hug-dev:
Currently only contains the part for Linux on x86, with a native
installation.

Fix #89

### DIFF
--- a/src/getting_started/README.md
+++ b/src/getting_started/README.md
@@ -1,0 +1,11 @@
+# Getting Started
+
+Quickly get the Parsec service up and running to try it out! Choose the flavour that you would
+prefer:
+
+- [Quickstart for Linux on x86](linux_x86.md): running Parsec with Mbed Crypto natively on your
+   machine, the fastest
+- [Quickstart using a Docker container](docker.md): running a Docker container with all providers
+   included, the most complete
+
+*Copyright 2021 Contributors to the Parsec project.*
--- a/src/getting_started/linux_x86.md
+++ b/src/getting_started/linux_x86.md
@@ -1,0 +1,106 @@
+# Quickstart for Linux on x86
+
+Run Parsec with the Mbed Crypto provider natively on your machine.
+
+## Overview
+
+In this guide you will learn how to quickly get started with Parsec in a toy environment. More
+specifically you will:
+
+- download and start the Parsec service
+- perform some Parsec commands using `parsec-tool`
+- see (and maybe run) Rust and Go clients code example
+
+Note that this installation method is only made for educational purposes and is not secure. To
+securely install Parsec on Linux for production, check [this guide
+instead](../parsec_service/install_parsec_linux.md).
+
+## Prerequisites
+
+- a 64 bit Linux running on x86
+- GLIBC version 2.27+ (check it with `ldd --version`)
+
+## Downloading Parsec
+
+Run the following command to download and unpack the `linux_x86` folder. It contains the Parsec
+service (`parsec`), its pre-built configuration (`config.toml`) and the Parsec tool (`parsec-tool`).
+
+```
+curl -s -N -L https://github.com/parallaxsecond/parsec/releases/download/0.7.0/quickstart-0.7.0-linux_x86.tar.gz | tar xz
+```
+
+## Starting Parsec
+
+From within the `quickstart-0.7.0-linux_x86` directory, start the Parsec service:
+
+```
+$ ./parsec &
+[INFO  parsec] Parsec started. Configuring the service...
+[INFO  parsec_service::key_info_managers::on_disk_manager] Found 0 mapping files
+[INFO  parsec_service::utils::service_builder] Creating a Mbed Crypto Provider.
+[INFO  parsec] Parsec is ready.
+```
+
+The Parsec service will now wait listening from clients connecting via `parsec.sock`.
+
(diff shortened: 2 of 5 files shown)


hug-dev:
To try this, added an archive to the `0.6.0` [release](https://github.com/parallaxsecond/parsec/releases/tag/0.6.0). It contains the current version of Parsec as opposed as the `0.6.0` one. Was only for testing.
If we agree with the name of the the archive, the various directoty, the steps and the guide then I can do this for the incoming `0.7.0` release instead.

Rust and Go examples can be added later (if needed). Can do a subsequent PR for the Docker part. It would only be running a docker container that contains everything in it, inside the user would only have to run `./parsec-tool` (or the examples) and it works.

hug-dev:
Links checked at #99

src/getting_started/README.md
```diff
@@ -0,0 +1,11 @@
+# Getting Started
+
+Quickly get the Parsec service up and running to try it out! Choose the flavour that you would
+prefer:
+
+- [Quickstart for Linux on x86](linux_x86.md): running Parsec with Mbed Crypto natively on your
+   machine, the fastest
+- [Quickstart using a Docker container](docker.md): running a Docker container with all providers
+   included, the mots complete
```
ionut-arm, line 9:
```suggestion
   included, the most complete
```

src/getting_started/linux_x86.md
```diff
@@ -0,0 +1,107 @@
...
+
+The commands should be run as a normal user (not root).
+
+## Downloading Parsec
+
+Run the following command to download and unpack the `linux_x86` folder. It contains the Parsec
+service (`parsec`), its pre-built configuration (`config.toml`) and the Parsec tool (`parsec-tool`).
+
+```
+curl -s -N -L https://github.com/parallaxsecond/parsec/releases/download/0.6.0/linux_x86.tar.gz | tar xz
```
ionut-arm, line 30: I think the folder inside the tarball should be called `linux_x86`, but I wouldn't call the file that since it will be tied to the version number. It might give the right impression to someone who looks at the release stuff alone, not related to this guide.

hug-dev: Will rename both with the `parsec-0.7.0-linux_x86` name as SPIFFE does so it's clearer.

ionut-arm: No, what I meant was that just saying something generic like that is not a good idea because someone might think that's the whole of `parsec`, packaged for Linux x86 systems (and it's not). The thing added to the release, at least, should be named more specifically, though a good name doesn't come to mind right now

hug-dev: ah I see your point. I can name it `quickstart-0.7.0-linux_x86`?

ionut-arm: Yes, I think that's good! maybe you can also throw in a `README` file that says what the package is for?

hug-dev:
Ok updated with your suggestions! Removed the asset from the `0.6.0` but added it to `0.7.0`.

src/getting_started/linux_x86.md
```diff
@@ -0,0 +1,106 @@
...
+- see (and maybe run) Rust and Go clients code example
+
+Note that this installation method is only made for educational purposes and is not secure. To
+securely install Parsec on Linux for production, check [this guide
+instead](../parsec_service/install_parsec_linux.md).
+
+## Prerequisites
+
+- a 64 bit Linux running on x86
+- GLIBC version 2.27+
```
ionut-arm, line 21: Maybe put a short comment in brackets to explain how to check it

ionut-arm (commented):
Looks good so far!

PR #105, opened 2021-03-26
Add instructions for local docs.rs builds

ionut-arm:

### DIFF
--- a/src/contributing/package_management.md
+++ b/src/contributing/package_management.md
@@ -39,11 +39,6 @@ to publish a crate, check out the [Dependency management](#dependency-management
 Once you are able to publish crates you can start doing so, but being aware of the following facts
 is helpful.
 
-- If the crate you are publishing is new or has major changes in its build steps, its documentation
-   might not be built properly by docs.rs. This is because docs.rs has its own build system where
-   crates are built in complete isolation from the internet and with a set of dependencies that
-   might not cover the ones you require. Check out their documentation
-   [here](https://docs.rs/about/builds).
 - Once you have settled on a crate to upstream, make sure you are on the most recent version of the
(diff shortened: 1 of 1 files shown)


src/contributing/package_management.md
```diff
@@ -57,6 +52,23 @@ is helpful.
...
+   might not cover the ones you require. Check out their documentation
+   [here](https://docs.rs/about/builds). If you want to be extra-careful and sure that your
+   documentation will build successfully, you can set up a local `docs.rs` and attempt to build your
+   crate. You can find the instructions
+   [here](https://github.com/rust-lang/docs.rs#getting-started). Follow them up to setting up the
+   external services, then run the command for building a local crate from
+   [here](https://github.com/rust-lang/docs.rs#build-subcommand). Be advised that setting up the
+   docs.rs build locally requires 10 GiB of disk space (or more) and is quite slow the first time
+   around, but caching helps with subsequent builds. It is therefore not recommended for CI runs,
+   though manually triggered builds might be ok. Also be advised that even if the build fails, the
```
ionut-arm, line 67: Do you reckon we should set up manual Github actions to check for this when we want to release?

hug-dev: Maybe having a worflow that is only manually triggered would be nice? So that you activate just before you are about to publish?

hug-dev (approved):
Thanks!

src/contributing/package_management.md
ionut-arm: Yeah, that was the question, whether it'd be helpful to have it for the crates where we do FFI (or have to link with specific libraries)

hug-dev: Yeah I think so! for crates where we only dynamically link to libraries, ie only the `tss-esapi` one for now

PR #108, opened 2021-04-01
Add a Parsec Stability page

### LINKED ISSUE #83: Create a chapter that defines the rules for stable releases of components
### Summary
Create a chapter in the book, which sets out the rules that must be followed to ensure that ongoing development is done in a backwards-compatible fashion from any given "stable" baseline release. These rules should enable the project to make a release that we can confidently deploy into long-term supported environments such as enterprise Linux distributions.

### Background
Parsec is a long-running service that will be configured, deployed, upgraded and maintained over a long period of time. The environment will also have client applications that are calling it, and persistent state that is collected on storage. Components will be upgraded over time. For Parsec to be supportable in enterprise system, it must not be brittle against upgrades. If Parsec is upgraded and restarted, client applications should continue working, and all persistent state from prior versions should remain valid and usable.

### What Is Needed?
Stability rules are primarily needed for the wire protocol and the service, since it is the Parsec service that will be packaged and deployed with Linux distributions. Client libraries will typically *not* be packaged as binaries, and will rather be consumed as source code dependencies, and their versioning systems will tend to be language-specific, with versioning rules (such as semantic versioning) being applied on a per-library basis. One grey area here is the `parsec-tool`, which, like the service, might be packaged as a binary, and might be upgraded separately from the service, while also being consumed by shell scripts on the system (which we don't want to break).

Ultimately, all of the rules will need to be enforced by CI tests so that PRs can automatically be checked, which would likely involve installing a baseline version, running a test suite, and then upgrading to the new version and running a regression suite of some kind. However, for now the scope is limited to capturing and documenting the rules. We can't enforce the rules until we have stated what they are.

Fortunately, much of Parsec has already been designed with backwards compatibility in mind, but some of these design choices need to be made explicit.

In no particular order, here's a non-exhaustive list of things that we may need to include. We can use comments on this issue to refine these and gather more ideas.

- Wire protocol versioning. The wire protocol has a versioning mechanism that must be strictly honoured. Request/response header fields must not be modified or re-ordered in anyway. The last breaking change to this was in May 2020 when we reworked it to align the request/response formats to make the protocol more amenable to shared memory transports. While it is possible to change the protocol, we must use the versioning mechanism, and we must negotiate with the client.
- API versioning. The API must strictly follow the open/closed principle. Contracts for published APIs must not be modified in a backwards-incompatible fashion. New opcodes may be added, but opcodes must never be removed. A provider does not need to support every opcode, however providers should not retract support for opcodes that were implemented in previous versions.
- Persistent state. Any persistent state (such as key properties stored on disk) must remain valid. If any changes of format are introduced, the service must be able to read the old format and upgrade to the new format on the fly. If this is not possible, then the new format must be by some kind of opt-in mechanism. For example, it could be a new type of key identity manager that needs to be selected in the service config.
- Configuration. Once a service has been deployed and configured, its configuration file (`config.toml`) must remain valid. If we add new features to the service, and those features require configuration entries, then either the configuration entries must be optional (with suitable in-code defaults or feature de-activation), or there must be an upgrade process that ensures new config entries are present *without* overwriting config that was already applied previously.

### Client Libraries
As mentioned above, client libraries are subject to somewhat different rules because they typically would not be deployed as binary packages, hence API compatibility is more important than ABI. For the case of the Rust and Go clients, this is largely covered by semantic versioning, and it becomes an issue for the individual clients and the per-language ecosystems in which they operate. The parsec-book would not be expected to cover all of these, except perhaps with a few general notes.

One open question for the client side is to what extent they should be calling `ListOpcodes` to determine support for particular operations. As and when Parsec reaches some "first stable release" milestone, is that the point where we decide that the opcode set is fixed, and clients should never assume the existence of any newer opcodes without first querying dynamically?

hug-dev:
This page describes what stability is for the Parsec service and how it is enforced/tested.

This is a work-in-progress page to put the framework in place.

I think the plan would be:
- create a `stability` label for issues
- for each one of the stability requirement, create a dedicated issue to make sure that stability is currently ensured in Parsec and add tests.
- add a link to the specific stability tests in the page per requirement

I think the ideal plan would be to have a stable Parsec version at the end of this, `0.X.0` for people/packagers to try out. After a few months of fixing stability problems, if they arise, we should then be confident enough to tag Parsec `1.0.0`.

I am still worried about the [C-STABLE](https://rust-lang.github.io/api-guidelines/necessities.html#c-stable) rule and not sure yet what it means for Parsec, a binary application. I think it would translate to the exact content of the stability document.

Please feel free to review the vocabulary I used and let's discuss it. For example the stability definition, "communication channel", "stability requirement", "stability enforcement", etc...
Feel free to indicate if I missed anything.

Fix #83


src/parsec_service/stability.md
```diff
@@ -0,0 +1,68 @@
...
+and restarted, client applications should continue working, and all persistent state from prior
+versions should remain valid and usable.
+
+## What does stability mean for the Parsec service?
+
+The following definition of stability is chosen for the Parsec service: *whatever two versions A and
+B of the Parsec binary, B being newer than A, A and B are stable if A can be removed and replaced by
+B without breaking anything*.
+
+Note that we are not looking at stability in the reverse direction: B can not be replaced by A
```
ionut-arm, line 20:
```suggestion
Note that we are not looking at stability in the reverse direction: B cannot be replaced by A
```

src/parsec_service/stability.md
```diff
@@ -0,0 +1,68 @@
...
+future stable version to that will be `1.x.y`.
+
+## What needs to be stable?
+
+With the above definition, Parsec stays stable if it keeps the exact same communication channels and
+format with its environment. Those are already described in the [dataflow
+diagram](../parsec_security/parsec_threat_model/threat_model.md#dataflow-diagram) in the threat
+model.
+
+The communication channels that needs to be stable are (some of them are not in the diagram):
```
ionut-arm, line 34:
```suggestion
The communication channels that need to be stable are (some of them are not in the diagram):
```

src/parsec_service/stability.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. Communication with clients
+2. Communication with authenticators
```
ionut-arm, line 37: Does this include the paths used to connect clients to service or service to authenticators?

ionut-arm: Also, do they cover the "logical" contents of the communication? For example, if one version of the service (or of a provider) accepts some set of parameters and another doesn't (or the other way around), is that a breaking change? Fixing a bug shouldn't really fall into "breaking changes", but how big of a bug can we allow that for?

The list keys bug is a good example (though a bit contrieved): in one version one key reports as belonging to `MbedCryptoProvider`, in the next as `TpmProvider` - is that a breaking change? I guess this isn't that controversial because the first one is clearly wrong, but I can imagine cases where something might be either treated as an error or just ignored, and we switch from one to the other.

src/parsec_service/stability.md
```diff
@@ -0,0 +1,68 @@
...
+| Configuration file             | Old configuration files should still work with future stable Parsec versions, with the same default for optional options.                              | Configuration options should not disappear in future stable Parsec versions. Configuration defaults should remain the same. New options should be optional.                                                                                                         | TODO           |
+| Key mappings                   | Old key mappings should still be read correctly by future stable Parsec versions.                                                                      | The way the KIM persistently stores the mappings need to be backward-compatible in regards with reading/writing.                                                                                                                                                    | TODO           |
```
ionut-arm, line 63: I guess the trickier question is how we actually handle breaking-change transitions here, but that's probably for another document.

src/parsec_service/stability.md
```diff
@@ -0,0 +1,68 @@
...
+The communication channels that needs to be stable are (some of them are not in the diagram):
+
+1. Communication with clients
+2. Communication with authenticators
+3. Communication received from the CLI invocation
+4. Configuration file (including default configuration options)
+5. Key mappings
+6. OS signals
+7. systemd communication
+8. Dynamic libraries dependencies
```
ionut-arm, line 43: I think there's another thing needed here: 
```
9. Provider workflow and call mapping
```
The name isn't the best, sorry. Imagine a key being created in the PKCS11 provider, restricted to some mechanism based on the key attributes sent in `PsaGenerateKey`. Then in the next version of the service, we make some changes - even with the same key attributes, we now create the key without a restriction on the mechanism, and instead specify it in some other way (when we make the call for signing, for example). Maybe this isn't allowed if the key already has those properties set by the previous version of the service, and the operation now fails. I don't know if the example works with PKCS11, it sounds vaguely familiar, but it's definitely a thing for TPMs.

I think this is tricky to define, especially because the number of options is huge. Maybe we can say that
1) objects created by the providers in their backends should be immutable
2) the way those immutable objects are created should not change between stable versions
???

ionut-arm (commented):
One other question I had was on the relationship between version numbers: the Parsec service version number, wire protocol version number, and provider/authenticator version numbers. Do we need to define how these relationships work and how they relate to the whole Stability goal?

paulhowardarm:
> One other question I had was on the relationship between version numbers: the Parsec service version number, wire protocol version number, and provider/authenticator version numbers. Do we need to define how these relationships work and how they relate to the whole Stability goal?

I think I would avoid getting into anything like that, because it has the potential to create an entirely unnecessary set of additional constraints on what is already quite a difficult problem. I am comfortable with letting different aspects of versioning drift and evolve at their own pace. I'd be wary of getting into situations where we need to artificially bump numbers or massage them to suit some over-arching schema.

paulhowardarm (commented):
Not wanting to inflict scope-creep on this PR, but somewhere along the line I think we need to address the story of how Parsec and Mbed TLS versioning intersect, both in the back-end (Mbed provider) and the front-end (C client using the SE driver). What are our goals here? As soon as there's an Mbed LTSR containing the PSA interfaces, should we expect Parsec to "just work" with whatever is installed in a distro, rather than grabbing specific versions as we do now? For the back-end, it would be nice to not have to link statically to Mbed Crypto, because it creates a maintenance burden within Parsec - we have to update to new versions to get patches for bugs/CVEs. For the front-end, the question is whether a client application could do something similar - just use whatever MBed Crypto is sitting on the target platform, and successfully register the SE driver. I don't think Parsec can solve this problem in isolation. It would have to be a dialogue with Mbed about what stability looks like on their side as well, both for API and ABI. But I can appreciate that this document might not be the place to try to address that question. It needs to be on the overall stability goals radar, however.

hug-dev:
> 
> 
> > One other question I had was on the relationship between version numbers: the Parsec service version number, wire protocol version number, and provider/authenticator version numbers. Do we need to define how these relationships work and how they relate to the whole Stability goal?
> 
> I think I would avoid getting into anything like that, because it has the potential to create an entirely unnecessary set of additional constraints on what is already quite a difficult problem. I am comfortable with letting different aspects of versioning drift and evolve at their own pace. I'd be wary of getting into situations where we need to artificially bump numbers or massage them to suit some over-arching schema.

I think this only defines the service version. As everything is linked together it would also be including the providers. Note that the core provider's version is the same as the service currently (the one in the Cargo.toml file). The wire protocol version is used as part of the stability in between the service and clients.

src/parsec_service/stability.md
hug-dev: > Does this include the paths used to connect clients to service or service to authenticators?

1 is for clients to service and 2 for service to authenticators.

> Also, do they cover the "logical" contents of the communication? For example, if one version of the service (or of a provider) accepts some set of parameters and another doesn't (or the other way around), is that a breaking change? Fixing a bug shouldn't really fall into "breaking changes", but how big of a bug can we allow that for?
>
> The list keys bug is a good example (though a bit contrieved): in one version one key reports as belonging to MbedCryptoProvider, in the next as TpmProvider - is that a breaking change? I guess this isn't that controversial because the first one is clearly wrong, but I can imagine cases where something might be either treated as an error or just ignored, and we switch from one to the other.

That's a good question. My first feeling would that this only covers the format of the communication: how the messages are sent and received but not their logical content. I would say that for the logical content clients can make stable runtime decisions without breaking if the content changes. For example we could not publish a new breaking change each time "ListOpcodes" returns one more opcode but clients can make decisions based on that that will be stable in the future.

Maybe the definition of stable ("without breaking anything") is a bit too loose as obviously a lot of things could break if `ListKeys` would only return "MbedCryptoProvider".

src/parsec_service/stability.md
hug-dev: Yes I will open a specific issue for each of those points where we can discuss that.

src/parsec_service/stability.md
hug-dev: I see what you mean and for me this is part of the contracts' definition in the wire protocol (so part of point 1).
By that I mean that if [PsaGenerateKey](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_generate_key.html) is modified so that the policies do not work in the same way, then there would need to be a new GenerateKey operation in the new behaviour.
For me the contract both describes the format of the operation (protobuf) and also its behaviour. Those two things should not break I think.

ionut-arm: It's not strictly about that, because you could reasonably implement the same operation that behaves the same way and correctly follows the Parsec frontend API, but uses the backend API in different ways, especially when it comes to the exact list of attributes we assign to the key. And this varies quite a lot from one backend to another.

Think about the checks for whether a key can be used to sign, within the TPM provider. In version 1.0.0 the checks might be done in software in the provider. If we then move, in version 1.1.0, to add the attributes on the key so that the TPM itself returns an error if someone tries to sign with a key that's not intended for signing, we can remove the software checks. All is fine, because both versions correctly implement the same interface and honor the contract in the same way. But a key created in 1.0.0 will not necessarily result in an error if it is misused, because we didn't set the correct attributes.

hug-dev:
> Not wanting to inflict scope-creep on this PR, but somewhere along the line I think we need to address the story of how Parsec and Mbed TLS versioning intersect, both in the back-end (Mbed provider) and the front-end (C client using the SE driver).

The stability document I wrote was only focusing on the Parsec service but not the clients. I thought that the semver rules where enough there to describe staiblity. I guess things are a bit more complicated for the SE driver, and we should maybe open a separate issue there.
This makes we think that we would also need one on the `parsec-tool`.

> What are our goals here? As soon as there's an Mbed LTSR containing the PSA interfaces, should we expect Parsec to "just work" with whatever is installed in a distro, rather than grabbing specific versions as we do now? For the back-end, it would be nice to not have to link statically to Mbed Crypto, because it creates a maintenance burden within Parsec - we have to update to new versions to get patches for bugs/CVEs.

For the Mbed Crypto provider, Parsec is relying on the `psa-crypto` crate. Currently, Mbed Crypto is compiled and linked statically during the build of this crate but if it is given the `MBEDTLS_LIB_DIR` and `MBEDTLS_INCLUDE_DIR` environment variables it will dynamically link with that instead. Ideally we would like something more robust like for the `tss-esapi` crate where `pkg-config` is used to find the Mbed Crypto library location so that no environment variable is needed.
Because the ABI of Mbed Crypto is not stable between versions, Parsec will have to be recompiled with the newer version of Mbed Crypto in order to generate the good mappings.

> For the front-end, the question is whether a client application could do something similar - just use whatever MBed Crypto is sitting on the target platform, and successfully register the SE driver. I don't think Parsec can solve this problem in isolation. It would have to be a dialogue with Mbed about what stability looks like on their side as well, both for API and ABI. But I can appreciate that this document might not be the place to try to address that question. It needs to be on the overall stability goals radar, however.

That is currently possible for the SE driver, it can be compiled by passing the `MBEDTLS_INCLUDE_DIR` environment variable poiting to the header of the current Mbed Crypto version used. Because of the similar ABI stability, it would have to be recompiled if the Mbed Crypto version changes.

src/parsec_service/stability.md
```diff
@@ -0,0 +1,68 @@
...
+diagram](../parsec_security/parsec_threat_model/threat_model.md#dataflow-diagram) in the threat
+model.
+
+The communication channels that needs to be stable are (some of them are not in the diagram):
+
+1. Communication with clients
+2. Communication with authenticators
+3. Communication received from the CLI invocation
+4. Configuration file (including default configuration options)
+5. Key mappings
```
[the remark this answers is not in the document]
hug-dev: Fair enough!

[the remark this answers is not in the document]
hug-dev: Actually, I think it is better to limit the document to what is there in Parsec today. If there is a need in the future for the service to store other kind of persistent data, then we should address the stability of that separately. The reason is that the methods we are going to use to store the key mappings in a stable way might be different to what we use for other kind of data.

src/parsec_service/stability.md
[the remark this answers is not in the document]
hug-dev: That's true! I guess I can put this as an exception on this note that although rollbacks are not considered, new clients should still be able to communicate with old versions of the Parsec service. So that this document considers only service upgrades for all communication channels except for (1) where both upgrades and downgrades are stable.

src/parsec_service/stability.md
```diff
@@ -0,0 +1,68 @@
...
+The following definition of stability is chosen for the Parsec service: *whatever two versions A and
+B of the Parsec binary, B being newer than A, A and B are stable if A can be removed and replaced by
+B without breaking anything*.
+
+Note that we are not looking at stability in the reverse direction: B can not be replaced by A
+without breaking anything.
+
+The principle of [semantic versioning](https://semver.org/) is used to describe the stability of
+Parsec versions and when breaking changes are done. If `parsec` version is at `1.0.0` then all
+future stable version to that will be `1.x.y`.
```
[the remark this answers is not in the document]
hug-dev: I agree that other than describing stability, being stable for a long time is the main goal here. With that paragraph I did not want to imply that using semver would be a reason for us to submit breaking changes often but more as a way to describe our stability. Even if it never happens (what is our goal), the semver definition allows us to prepare for that eventuality.
Moreover I would say that it would be a good sign for stakeholders to see that Parsec is still at `1.x.y` after a very long time.

I will add a clarification sentence that staying stable in the same `1.x.y` is indeed the goal here.

src/parsec_service/stability.md
hug-dev: Thanks for the precision and the example.

I still thinks it's related with respecting the contract (so point 1): no matter at what version a key is created (and then used), it should respect the contract as described in the key generation operation. If we introduce a bug where the contract is not longer respected when a key created in 1.0.0 is used in 1.1.0 then the stability of the communication channel with clients has been broken.

That is a good point though I did not think about. I can add a sentence about operation contract being stable in point 1. Testing that might need some new things that we currently do not have. For keys for example, their policies should be respected at at Parsec version.

ionut-arm: I think it's a bit of a stretch to include it in 1 - the same could be said about `5. Key mappings`, that no matter at what version a key mapping is created, it should allow providers to respect their contracts. My example was restricted to a change in behaviour, but it could very well be that some changes in the way we use the backend leaves all old keys completely unusable.

It would indeed be very nice if we could test, and I'm sure we can. For all of them we have the keys and all the configuration stored on disk, there should be a way to preserve an environment with a bunch of keys provisioned and test in it. This would also cover the testing for key mappings. I'll raise an issue in the service repo

hug-dev: I agree that 1 then becomes big but I still think that the respect of the contract is included in the communication channel with the client. The contract is what the client assumes Parsec will respect. I can split 1 in 3 parts to make it more readable/easier:
- definition of the format of requests/responses (ie the format)
- definition of the operation contracts (ie the logic)
- definition of the listener endpoints

About testing in general for this document, my plan was, once it is agreed, to create one single specific issue for each line of the table where we will look at how to stabilise and how to test.

src/parsec_service/stability.md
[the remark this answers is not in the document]
paulhowardarm: I'm not sure that I agree - this feels like letting details get in the way of the goal. If the aim of this stability document is to describe what we mean by stability for the service, then I think we need scope that in terms of all possible persistent state that might be stored and read back later. This doesn't preclude having a clear statement of how this would be achieved for individual pieces like the key mappings - I just think that we need a very strong top-level statement that covers all avenues.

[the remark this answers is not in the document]
hug-dev: > then I think we need scope that in terms of all possible persistent state that might be stored and read back later.

I see your thinking. I also think that this document should describe what we mean by stability for the service. I guess my methodology here was to list all current and used communication channels as opposed as the ones we could/will have in the future. That way for each of them we can actually link a specific issue to do the work to make sure stability is enforced and tested.

I am happy to modify this item as you suggest though: a top-level "Internally stored persistent states" statement and then indivual items like key mappings and the ones we add in the future.

paulhowardarm (approved):

ionut-arm (approved):

hug-dev:
Im creating the 10 specific issues and will make a PR with their link here after.

PR #110, opened 2021-04-14
Fix and add information around policies

hug-dev:
Specifically for signing/verifying hashes/messages.

### DIFF
--- a/src/parsec_client/operations/psa_key_attributes.md
+++ b/src/parsec_client/operations/psa_key_attributes.md
@@ -256,4 +256,55 @@ Definition of the usage flags. They encode what kind of operations are permitted
 | `verify_hash`    | Boolean | Permission to verify a hash with the key           |
 | `derive`         | Boolean | Permission to derive other keys from this key      |
 
+### sign message usage
+
+Permission to sign a message with the key.
+
+This flag allows the key to be used for a MAC calculation operation or for an asymmetric message
+signature operation, if otherwise permitted by the key’s type and policy. The flag must be present
+on keys used with the following APIs:
+
+- PsaMacCompute
+- PsaSignMessage
+
+For a key pair, this concerns the private key.
+
+### verify message usage
+
+Permission to verify a message signature with the key.
+
+This flag allows the key to be used for a MAC verification operation or for an asymmetric message
+signature verification operation, if otherwise permitted by the key’s type and policy. The flag
+must be present on keys used with the following APIs:
+
+- PsaMacVerify
+- PsaVerifyMessage
+
+For a key pair, this concerns the public key.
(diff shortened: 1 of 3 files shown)


src/parsec_client/operations/psa_key_attributes.md
```diff
@@ -256,4 +256,55 @@ Definition of the usage flags. They encode what kind of operations are permitted
...
+
+### sign hash usage
+
+Permission to sign a message hash with the key.
+
+This flag allows the key to be used to sign a message hash as part of an asymmetric signature
+operation, if otherwise permitted by the key’s type and policy. The flag must be present on keys
+used when calling PsaSignHash.
+
+This flag automatically sets `sign_message`: if an application sets the flag `sign_hash` when
```
hug-dev, line 293: This and next paragraph might mean that we need to change/adapt our checks in the Parsec interface. Either modify the checks for `sign_message` or have private fields on the usage flags and setters that do the correct thing.

ionut-arm: So do we do the fixes in the service first before merging this, or is it irrelevant between releases?

Also, on a similar note, did we ever decide what happens with "versions" of the book? I mean, some parts in the book might change in a way that isn't consistent with version 0.6.0 of the service. How would someone know what the book looked like back then, and how it changed since?

ionut-arm (approved):

src/parsec_client/operations/psa_key_attributes.md
hug-dev: > So do we do the fixes in the service first before merging this, or is it irrelevant between releases?

I think it's ok to merge this but we should create the relevent issues on the various repos. I think mostly `psa-crypto` and then import the changes upwards.

> Also, on a similar note, did we ever decide what happens with "versions" of the book? I mean, some parts in the book might change in a way that isn't consistent with version 0.6.0 of the service. How would someone know what the book looked like back then, and how it changed since?

Good point, not sure if we want to have a mapping between service version and book version but that sounds like a good idea. I guess for now they can still browse the git history 🤡 We only publish the `main` branch of the book anyway...

ionut-arm: > We only publish the main branch of the book anyway...

Yeah, I know, but someone would have to go through the release history of the service, align it with the changes here and then rebuild it from source. I suggest creating releases of the book for some milestones (e.g. version 0.7.0 of the service) and attaching a built version of the book. That can just be downloaded, unzipped and opened in a browser from disk.

An unfortunate issue is that we track both service and clients in the book (and much more, actually), so we have to decide what we create new "releases" for.

PR #115, opened 2021-05-04
Create a first attempt at a documented release process for Parsec.

paulhowardarm:
Signed-off-by: Paul Howard <paul.howard@arm.com>

### DIFF
--- a/src/contributing/release_process.md
+++ b/src/contributing/release_process.md
@@ -10,6 +10,12 @@ Precise dates are agreed by the maintainers and the project community on each cy
 cadence helps with the process of packaging Parsec releases into Operating System distributions and
 other downstream consumers.
 
+This page focuses on the process for twice-yearly major releases. Naturally, it is expected that
+minor point releases will also take place to address defects or to add other improvements that are
+needed urgently and cannot wait for the next major cycle. A fixed process for such point releases
+has not yet been established, but this will be determined according to which practices appear to
+work well within the project community.
+
 ## Responsibility for Releases
 
 It is the collective responsibility of the [Parsec project
@@ -47,6 +53,12 @@ Both of these components are included because they are both built from Rust and 
 of dependencies. A coordinated release allows for the versions of the shared Rust crates to be kept
 synchronized. These two components are also both commonly packaged for Operating System
 distributions.
+
+Additionally, each release includes a convenient
+[quickstart](https://parallaxsecond.github.io/parsec-book/getting_started/linux_x86.html) package,
+which contains the binaries for just the Parsec service and the Parsec tool, along with a basic
+configuration file and a README document with a link to the quickstart guide. This is to help new
+users gain some early familiarity with the system.
 
 Other components of the Parsec project, such as documentation and client libraries in languages
 other than Rust, are currently not included in the scheduled release cadence. These components are
@@ -153,12 +165,14 @@ weekly meetings. Please see the [community](https://github.com/parallaxsecond/co
 for more details. The project maintainers will be available to advise you on whether to use a
 feature branch.
 
-When using feature branches, please ensure that commits are reviewed by project maintainers at the
-time of submission to the feature branch. Do not wait until the entire feature is ready to be
-promoted into the main branch before requesting code reviews, since large and bulky code reviews are
-generally not well received by the maintainers, and this might delay the promotion of the changes.
+When using feature branches, please ensure that commits are reviewed by the project maintainers
+through the pull request process at suitable intervals, so that reviews can be kept small and
+manageable. Do not wait until the entire feature is complete before requesting code reviews, since
+large and bulky code reviews are generally not well received by the maintainers, and this might
+delay the approval of changes.
 
-To keep the repositories tidy, please delete feature branches or other temporary branches as soon as
-they are no longer needed.
+To keep the repositories tidy, the maintainers will aim to delete feature branches as soon as
+possible after the code changes have been promoted into the mainline. Please contact the maintainers
+through the community channels to request the deletion of a feature branch.
 
 *Copyright 2021 Contributors to the Parsec project.*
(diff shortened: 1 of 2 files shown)


src/contributing/release_process.md
```diff
@@ -0,0 +1,164 @@
...
+for more details. The project maintainers will be available to advise you on whether to use a
+feature branch.
+
+When using feature branches, please ensure that commits are reviewed by project maintainers at the
+time of submission to the feature branch. Do not wait until the entire feature is ready to be
+promoted into the main branch before requesting code reviews, since large and bulky code reviews are
+generally not well received by the maintainers, and this might delay the promotion of the changes.
+
+To keep the repositories tidy, please delete feature branches or other temporary branches as soon as
+they are no longer needed.
```
hug-dev, line 162: I think this is something only the maintainers can do, but the creator of a feature branch could inform the maintainer that it is ready to be deleted.

src/contributing/release_process.md
```diff
@@ -0,0 +1,164 @@
...
+If you are planning to contribute a significant feature or major architectural change, please
+contact the maintainers via the available community channels, such as the Slack channel or the
+weekly meetings. Please see the [community](https://github.com/parallaxsecond/community) repository
+for more details. The project maintainers will be available to advise you on whether to use a
+feature branch.
+
+When using feature branches, please ensure that commits are reviewed by project maintainers at the
+time of submission to the feature branch. Do not wait until the entire feature is ready to be
+promoted into the main branch before requesting code reviews, since large and bulky code reviews are
+generally not well received by the maintainers, and this might delay the promotion of the changes.
```
hug-dev, line 159: I think this is mandotory anyway as pushing code to a feature branch has to be done through a pull-request, just like pushing code to `main`.

src/contributing/release_process.md
```diff
@@ -0,0 +1,164 @@
...
+Code freeze dates may also vary depending on the content of each release. Typically, a code freeze
+will occur at around 2-4 weeks before the target release date.
+
+During the code freeze period, code submissions will only be accepted if they address critical
+defects or other issues that are blocking the release.
+
+Shortly after the code freeze date, a release candidate will be published for each of the Parsec
+service and Parsec tool. The availability of this release candidate will be communicated to
+downstream consumers such as Operating System maintainers, giving them the opportunity to perform
+integration testing before the release is finalized.
```
hug-dev, line 107: Just to be clear, I think it's fine for the release candidate to have a "real" semantic version, like `0.8.0`. If any bugs are discovered during the code freeze period, patch versions can be released subsequently (`0.8.1`, `0.8.2`, etc..). Hopefully, it won't be needed to add a new feature during the code feeze period.

This is to not have to use `-rc*` tags which can be messy with Cargo use.

src/contributing/release_process.md
```diff
@@ -0,0 +1,164 @@
...
+dashboard](https://github.com/orgs/parallaxsecond/projects/1) can be used to track the items that
+are being prepared for the next release. Use the
+[milestones](https://github.com/parallaxsecond/parsec/milestones) to filter the dashboard view in
+order to see the issue tickets that have been targeted for each release.
+
+Please be aware that the content of a release cannot be guaranteed. In order to ensure a fixed
+release schedule, it is sometimes necessary to modify the scope of each release. The project
+dashboard always provides the best available indication of the current plan.
+
+## Released Components
```
hug-dev, line 34: If we wanted to make it official, I think it would also be nice to indicate that a quickstart archive is also released as a GitHub Asset (see [here](https://github.com/parallaxsecond/parsec/releases/tag/0.7.0) at the bottom for example) with each version. This is to work with the [quickstart](https://parallaxsecond.github.io/parsec-book/getting_started/linux_x86.html) guide.

hug-dev (commented):
Thanks it looks good!!
Adding to what I wrote on the page I have two other general comments:

1. Major releases are going to be cadenced but it is maybe worth it to indicate that patch releases are possible in between to fix (security) bugs? In some cases we should maybe not wait 6 months before releasing a version with a fix if the bug is bad.
2. I guess we are not detailing any process at this stage for *version branches*? ie branching from `main` from the code freeze date and putting there all bug fixes before the release? Would also be needed long-term for patching of *old* Parsec versions if Parsec can not be upgraded in some environments. It's probably fine to only worry about that when a problem comes.

src/contributing/release_process.md
paulhowardarm: Yes, good point - will reword this,

src/contributing/release_process.md
paulhowardarm: So maybe what we need to be saying here is that people who are using feature branches should ensure that they are actual branches and not personal forks, where arbitrary amounts of work could get done without any visibility from the maintainers?

src/contributing/release_process.md
paulhowardarm: This is the kind of detail I'm intentionally glossing over for now, because I would like to see how the "practice run" for this process works out later in the year. Once we've done that first try, I think we'll probably come back to this page and add quite a few specific details and refinements.

src/contributing/release_process.md
hug-dev: Yes I think so! But maybe actually this could be covered with just removing some words:

```diff
-When using feature branches, please ensure that commits are reviewed by project maintainers at the
time of submission to the feature branch.
-Do not wait until the entire feature is ready to be
promoted into the main branch before requesting code reviews, since large and bulky code reviews are
generally not well received by the maintainers, and this might delay the promotion of the changes.
+When using feature branches, please ensure that commits are reviewed by project maintainers regularly.
+Do not wait until the entire feature is ready before requesting code reviews, since large and bulky code reviews are
generally not well received by the maintainers, and this might delay the promotion of the changes.
```

hug-dev (approved):
Thanks for adding the changes!! Looks very good to me now 👌

ionut-arm (approved):
Looks good!

PR #118, opened 2021-05-12
Add testing policy

### LINKED ISSUE #117: Publish testing guidelines
We currently have coding guidelines and information on how to run tests, and what those tests do, but no real guidelines for what tests to add when new functionality is introduced.

These testing guidelines should refer to the types of tests we already support and explain, plainly, what kind of tests are needed for various types of features, and how extensive these tests should be.

ionut-arm:
This commit adds a policy describing the types of tests expected for new
functionality added in a PR.

Fixes #117

### DIFF
--- a/src/contributing/pr_checklist.md
+++ b/src/contributing/pr_checklist.md
@@ -66,4 +66,46 @@ If new kinds of tests are added, please check the
 [testing](https://parallaxsecond.github.io/parsec-book/parsec_service/tests/index.html) page and its
 child.
 
+## Testing
+
+Pull requests introducing new features create new ways in which the service interacts with its
+environment. All such new avenues must be thoroughly tested to check both for potential flaws in the
+current implementation, and possible regressions due to future iterations or changes.
+
+A list of current types of tests can be found
+[here](https://parallaxsecond.github.io/parsec-book/parsec_service/tests/existing_tests.html). The
+list only looks at the high-level types of tests that we support, but should be a good guide to what
+types of tests must be enabled or written. For example:
+
+- When a **new provider** is introduced, or when **support for some existing operations is added to
+   an existing provider**, make sure all the end-to-end tests that cover the new functionality are
+   enabled. This is generally done by means of Cargo features in the `e2e-tests` crate. If the
+   provider doesn't support the functionality needed for some test, make sure it is disabled using
+   such features, e.g. if only EC keys are supported for asymmetric signatures, tests based on RSA
+   keys should be explicitly disabled (i.e. `#[cfg(not(feature = "some-provider"))]`), as we
+   currently have no mechanism for runtime detection of supported parameters. If the provider
+   includes functionality which does not have proper testing (e.g. a specific key type, an elliptic
+   curve, etc.), tests must be added. Ideally, when creating a new test for cryptographic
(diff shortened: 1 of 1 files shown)


src/contributing/pr_checklist.md
```diff
@@ -66,4 +66,45 @@ If new kinds of tests are added, please check the
...
+   interaction model with key info managers, to ensure stability in the long run. As such, please
+   ensure that the provider is set up in key mappings tests which take care of stability of data
+   flows between providers and data stores.
+- When a **new authenticator** is introduced, make sure tests that mimic the full environment for a
+   number of clients are introduced. The tests should verify that clients are properly separated
+   (they cannot access each other's objects), that clients without permissions (if that concept
+   exists for the authenticator) cannot access the parts of the service which require
+   authentication, and that admins are recognized properly. Some of these checks can be implemented
+   as unit tests.
+- When a **new operation** is introduced into the service, testing must ensure not only that the
```
hug-dev, line 99: Maybe parts of this can be shared with the first point for clarity? As in doing one point "when a new provider is introduced" and another "when a new operation is added in one provider".

src/contributing/pr_checklist.md
```diff
@@ -66,4 +66,45 @@ If new kinds of tests are added, please check the
...
+[here](https://parallaxsecond.github.io/parsec-book/parsec_service/tests/existing_tests.html). The
+list only looks at the high-level types of tests that we support, but should be a good guide to what
+types of tests must be enabled or written. For example:
+
+- When a **new provider** is introduced, or when support for some **new operations is added to an
+   existing provider**, make sure all the end-to-end tests that cover the new functionality are
+   enabled. This is generally done by means of Cargo features in the `e2e-tests` crate. If the
+   provider doesn't support the functionality needed for some test, make sure it is disabled using
+   such features, e.g. if only EC keys are supported for asymmetric signatures, tests based on RSA
+   keys should be explicitly disabled. If the provider includes functionality which does not have
```
hug-dev, line 85: Maybe say how the test can be disabled? As in it's better to disable it at runtime (as check in the operation is supported or not) rather than not compiling it.

hug-dev (approved):
Thanks!

src/contributing/pr_checklist.md
ionut-arm: Hmm, yeah, though the first point is somewhat different: "when an existing operation is added to an existing provider that didn't have that operation yet" is more precise, but sounds like a complete drag...

hug-dev: I guess "when an existing operation is added to an existing provider that didn't have that operation yet" would cover all cases? As when you introduce a new operation you would add it to the specific providers that would support it after.

ionut-arm: Well, technically yes, but take for example my PR for adding asymmetric encryption support to the TS provider - I don't have to add all the asymmetric encryption tests. If this was the first time the operation was introduced in the service, I would need to write those tests.

src/contributing/pr_checklist.md
ionut-arm: Attempted a fix - does it sound more reasonable now?

hug-dev (approved):
Yep, it's good like that!

PR #120, opened 2021-05-14
Add paragraph about non-breaking changes

hug-dev:
Adds some words about allowing non-breaking changes under certain constraints.

In relation with https://github.com/parallaxsecond/parsec/issues/388


src/parsec_client/api_overview.md
```diff
@@ -88,9 +88,29 @@ below.
...
+In some cases, it might be helpful to update contracts in a *non-breaking way*. If multiple
+operations share the same parameters (such as the PSA algorithm or key attributes), this avoids
+having to duplicate all of them if an update is done to those parameters. To make sure that backward
+compatibility with client code is always preserved and that the Parsec service remains stable
+through updates, the allowed changes are restricted to the following:
+
+- Adding a new *optional* field to an existing type. If not set, the contract containing the new
+   field should behave exactly the same as the version without it. The new field should be ignored
+   if not recognised by an old version. That means that the functionality brought by a new optional
+   field might not be exerced if the service is older than the client, and that without an error
```
ionut-arm, line 105:
```suggestion
   field might not be exercised if the service is older than the client, without an error
```
?

ionut-arm: but now I have questions - how does the client know the service didn't see that field, if capability discovery isn't a thing?

hug-dev:
> but now I have questions - how does the client know the service didn't see that field, if capability discovery isn't a thing?

On the request itself it won't be able to (no error is returned). Depending on the contract, it can:
- check before the request with capabilities discovery (but not yet implemented)
- check after the request with `ListKeys`, if we are talking about a new field inside the KeyAttributes structure

Someone adding an optional field to an operation would have to keep that in mind, that the field might be ignored by the service. Depending on the field's nature, that might not be wanted in which case creating a new contract would be better.

ionut-arm (approved):

ionut-arm:
> check after the request with ListKeys, if we are talking about a new field inside the KeyAttributes structure

While I sort of understand this approach, it suddenly places the burden on clients to know which parameters might be discarded silently by the service... I get that there's no way to allow for non-breaking changes to be introduced otherwise, but it seems like such a horrible UX from the client's perspective: "Oh, so if I set _this_ particular flag I then have to always check that the service actually cared about it, whereas with all these other flags I don't."

hug-dev:
I think it should be clear in the documentation of the optional fields that this behaviour is possible... Agree this is not the best.

PR #124, opened 2021-07-23
Add key attestation operations

ionut-arm:
This commit adds documentation for two new operations: AttestKey, used
for generating proofs that some key provisioned through Parsec is indeed
by the claimed hardware; GetMakeCredParams, used to prepare the
parameters for one of the mechanisms used within AttestKey (namely,
ActivateCredential in the TPM provider).

### DIFF
--- a/src/parsec_client/operations/attest_key.md
+++ b/src/parsec_client/operations/attest_key.md
@@ -1,0 +1,60 @@
+# AttestKey
+
+Attest that a Parsec-managed key is protected by a hardware backend. Opcode: 30 (`0x001E`)
+
+**(EXPERIMENTAL) This operation is in an experimental phase. No guarantees are offered around the
+stability of the contracts and abstract definition of the operation, or of any associated key
+attestation mechanism.**
+
+## Parameters
+
+| Name                 | Type                                                 | Description                               |
+|----------------------|------------------------------------------------------|-------------------------------------------|
+| `attested_key_name`  | String                                               | Name of the key to attest                 |
+| `parameters`         | [`AttestationMechanismParams`](attest_key_params.md) | Attestation mechanism-specific parameters |
+| `attesting_key_name` | String                                               | Name of the key to use for attesting      |
+
+- The exact usage flags required by the attesting key are determined by the mechanism used, also
+   described on the [key attestation parameters page](attest_key_params.md).
+
+## Results
+
+| Name     | Type                                        | Description                           |
+|----------|---------------------------------------------|---------------------------------------|
+| `output` | [`AttestationOutput`](attest_key_params.md) | Attestation mechanism-specific output |
+
+## Specific response status codes
+
+TBD
+
+## Description
+
+This operation performs a key attestation using a mechanism supported by the backend holding the
+key. The purpose of the operation is to help a Parsec client provide proof to a 3rd party that some
+key provisioned by the client is indeed stored and secured by a hardware backend. As such, the
+operation is backed by native functionality in the hardware to attest to ownership of the key.
+
+Given the wide variety of possible mechanisms, many of the properties, restrictions, and formats
+involved are mechanism-dependent, including:
+
+- Properties of the attested key (e.g. whether it was created within the backend or imported)
+- Properties of the attesting key (e.g. if it must be able to sign or decrypt)
+- Number, content, and purpose of parameters required by the attestation
+- Contents and format of the output
+- Whether or not the attesting key can be specified
+
+All such characteristics are thoroughly described per mechanism on the [key attestation parameters
+page](attest_key_params.md).
+
+All instances of backends that support `AttestKey` must be configured with a default, root key that
+has been pre-provisioned and which can be used to produce attestations. This default attesting key
+is selected by leaving the `attesting_key_name` empty. If the backend allows other keys to be used
+for attesting, attestation chains can be created starting from the root key.
+
+`AttestKey` applies to asymmetric key pairs only.
+
+## Contract
+
+TBD
+
+*Copyright 2021 Contributors to the Parsec project.*
--- a/src/parsec_client/operations/attest_key_params.md
+++ b/src/parsec_client/operations/attest_key_params.md
@@ -1,0 +1,66 @@
+# Key Attestation Parameters
+
+This page defines the mechanism-specific inputs and outputs of `AttestKey`. For an in-depth look at
+the mechanisms and hardware tokens that we've considered, you can read our write-up
+[here](https://drive.google.com/file/d/11jLLv8zN_xRunj9BEkk_L_tj5DiQOtkG/view?usp=sharing).
+
+Each mechanism comes with its own definitions for `AttestationMechanismParams` and
+`AttestationOutput`.
+
+**(EXPERIMENTAL) The parameters for key attestation are in an experimental phase. No guarantees are
+offered around the stability of the interface for any key attestation mechanism.**
+
+## Activate Credential (TPM provider) {#activatecredential}
+
+The [TPM 2.0 Commands
+spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part3_Commands_pub.pdf)
+describes the `TPM2_ActivateCredential` operation as follows:
+
+> This command enables the association of a credential with an object in a way that ensures that the
+> TPM has validated the parameters of the credentialed object.
+
+`TPM2_ActivateCredential` allows a 3rd party to be assured of the protection of a key by means of an
+encrypted credential. The 3rd party produces a random credential and encrypts it using the algorithm
+defined in the [TPM 2.0 Architecture
+spec](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part1_Architecture_pub.pdf),
+section B.10.4. The outputs of that algorithm (the encrypted and HMAC-protected credential, and a
+secret seed encrypted with the public part of the attesting key) are sent to the Parsec service
+which proceeds to perform the operation and returns the decrypted credential. The 3rd party can then
+be certain that the key is protected by a TPM by confirming that the credential sent and the one
+received are identical.
+
+The computation mentioned previously relies on a number of parameters that must be obtained from the
+Parsec service. As some of these parameters are strictly TPM-specific, they can be retrieved with
+the [`PrepareKeyAttestation`](prepare_key_attestation.md) operation. You can see how to perform the
+preparation step for `ActivateCredential`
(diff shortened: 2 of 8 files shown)


src/parsec_client/operations/attest_key.md
```diff
@@ -0,0 +1,111 @@
...
+All such characteristics are thoroughly described per mechanism in the section below.
+
+All instances of backends that support `AttestKey` must be configured with a default, root key that
+has been pre-provisioned and which can be used to produce attestations. This default attesting key
+is selected by leaving the `attesting_key_name` empty. If the backend allows other keys to be used
+for attesting, attestation chains can be created starting from the root key.
+
+`AttestKey` applies to asymmetric key pairs only.
+
+## Attestation mechanisms
```
hug-dev, line 52: For the attestation mechanisms types, since it seems to be an enumerated type, it might be good to follow the same idea than for the [`KeyType`](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_key_attributes.html#keytype-type) type:
- separate it in its own file
- describe each `AttestationMechanismParams` and `AttestationOutput` as enumerated types: "A AttestationMechanismParams type can contain one of the following attestation mechanism: * ActivateCredential * ..."
- have a subsection where the ActivateCredential type is explained as it is done here but doing it in high-level term, without specifying the contract. Similarly as what we do for `EccKeyPair` with "Uses one of the curve family supported".

hug-dev (approved):

src/parsec_client/operations/attest_key.md
ionut-arm: I agree with the first two points, I was considering doing it like that, but wasn't sure about adding another file. 

> doing it in high-level term, without specifying the contract.

Not sure I follow what this means. Not to describe how the operation works and only focus on the formats/expected outputs?

hug-dev: > Not sure I follow what this means. Not to describe how the operation works and only focus on the formats/expected outputs?

I was comparing it with some of our enumerated types where we don't exactly show the contract but only "Uses one of the ... supported". And let the implementors of those contracts in the various languages (currently only protobuf) interpret that how they want. But I guess in the case of those types, the message can be written formaly as you did because the parameters are not just enums.

ionut-arm: I guess that's the reason - for those types you can say "it can be one of these values", but for `ActivateCredential` the contents are more closely resembling `PsaExportPublicKey`, so we have to define what the format of the data is, you can't leave it to the different contracts. 

I guess my question is whether the big description of the mechanism is useful in this document?

hug-dev: > I guess my question is whether the big description of the mechanism is useful in this document?

If we were to put the mechanisms in a separate document, I think it would make sense to be included there.

hug-dev (approved):
Nice, thanks for the changes!

src/parsec_client/operations/attest_key_params.md
```diff
@@ -0,0 +1,64 @@
+# Key Attestation Parameters
+
+This page defines the mechanism-specific inputs and outputs of `AttestKey`. For an in-depth look at
+the mechanisms and hardware tokens that we've considered, you can read our write-up
+[here](https://drive.google.com/file/d/11jLLv8zN_xRunj9BEkk_L_tj5DiQOtkG/view?usp=sharing).
+
+Each mechanism comes with its own definitions for `AttestationMechanismParams` and
```
paulhowardarm, line 7: I want to make sure I understand what is meant by "each mechanism comes with its own definitions". Do we really mean definitions? Or just interpretations of the fields? There can only be one definition of the structure in terms of the Protobuf messages. Will the two fields `credential_blob` and `secret` be common, but with different interpretations as to their meaning? Or are these two fields inherently specific to the TPM mechanism only? If it's the latter, then how do we expect Protobuf definitions to look when multiple mechanism are supported?

src/parsec_client/operations/get_make_cred_params.md
```diff
@@ -0,0 +1,60 @@
+# GetMakeCredParams
+
+Get the parameters for performing a `TPM2_MakeCredential` operation. Opcode: 31 (`0x001F`)
+
+**This is a TPM-specific operation.**
```
paulhowardarm, line 5: If this has to be a TPM-specific operation, then maybe our opcode naming needs a convention for this, maybe even  making it an opcode name prefix like `TpmGetMakeGredParam`, because this would be the first case where the opcode is entirely specific to the provider. But I wonder whether there is a way to avoid making it explicitly TPM-specific. It feels slightly lopsided to me that `AttestKey` is a generic operation that can accept a TPM-specific flavour of input, whereas `GetMakeCredParam` is an entirely TPM-specific opcode, rather than a generic opcode that can return a TPM-specific flavour of output. Either way, we know that the ultimate client of these APIs will need to have TPM intelligence, and that's okay. But is there a way to make the design more consistent in terms of how that intelligence is exercised? At the moment it looks a bit half-and-half. For `AttestKey` you need to make sure you pass in TPM-oriented parameters, but for `GetMakeCredParams` the whole opcode is TPM-specific.

I'm inclined to suggest that we do one of two things, we either make _both_ of these operations TPM-specific (`TpmAttestKey` and `TpmGetMakeCredParams`), and just dismiss any idea that attestation can ever be provider-neutral. Or, alternatively, remodel `GetMakeCredParams` as something like `PrepareKeyAttestation`, and have a TPM-specific interpretation of the results, which would be precisely analogous to how `AttestKey` has a TPM-specific interpretation of its parameters.

paulhowardarm (changes requested):
I have comments/suggestions which may or may not be helpful. The only change I would actually request is that we adopt a naming convention for opcodes that are fundamentally provider-specific, similar to the `Psa` prefix that we use for all PSA opcodes.

src/parsec_client/operations/attest_key_params.md
ionut-arm: I do mean that each mechanisms essentially defines a completely new set of fields. You can see how I'm thinking of implementing the protobuf contracts in section 3 of [the investigation doc](https://docs.google.com/document/d/11jLLv8zN_xRunj9BEkk_L_tj5DiQOtkG/edit?usp=sharing&ouid=110488261398236555557&rtpof=true&sd=true). We'd use the `oneof` protobuf feature that allows you to define a field essentially as an "enum" of different message types. So for `ActivateCredential` you get a "nested" message with those two fields (`credential_blob` and `secret`), for others you'd get something else. There are multiple advantages to doing it this way, including this flexibility/liberty of adding new definitions easily and in a backwards compatible manner.

src/parsec_client/operations/get_make_cred_params.md
ionut-arm: Do you have any preference between the two? I'll have to have a think about what either of them mean in terms of developer experience

ionut-arm: I think both options have an almost identical "footprint" for a client - they'd still need to use mechanism-specific code in both cases, just that in one of them (where both operations are TPM-specific) the client function call would be TPM-specific as well, not just the parameters that get passed to it.

For the service, the downside of generic operations is that providers would need to reject mechanisms they don't support, but would otherwise be cleaner to implement. I'm leaning towards the generic approach since it also suits Parsec goals better.

ionut-arm: Ok, I've made the 2nd operation generic as well, and wrote up the parameters/output of that in another file, much like for `AttestKey`.

src/parsec_client/operations/service_api_coverage.md
```diff
@@ -7,36 +7,38 @@ table.
...
+| [PsaSignHash](psa_sign_hash.md)                                  | ❎  | ✅         | ✅     | ✅     | ✅             | ❌                |
+| [PsaVerifyHash](psa_verify_hash.md)                              | ❎  | ✅         | ✅     | ✅     | ✅             | ❌                |
+| [PsaAsymmetricEncrypt](psa_asymmetric_encrypt.md)                | ❎  | ✅         | ✅     | ✅     | ❌             | ❌                |
+| [PsaAsymmetricDecrypt](psa_asymmetric_decrypt.md)                | ❎  | ✅         | ✅     | ✅     | ❌             | ❌                |
+| [PsaRawKeyAgreement](psa_raw_key_agreement.md)                   | ❎  | ✅         | ❌     | ❌     | ❌             | ❌                |
+| [PsaGenerateRandom](psa_generate_random.md)                      | ❎  | ✅         | ❌     | ❌     | ❌             | ✅                |
+| [AttestKey](attest_key.md)                                       | ❌  | ❌         | ❌     | ❌     | ❌             | ❌                |
+| [PrepareKeyAttestationParams](prepare_key_attestation_params.md) | ❎  | ❎         | ❎     | ❌     | ❎             | ❎                |
```
hug-dev, line 41: Should that be `PrepareKeyAttestation`? And now that it is generic, I guess it could be implemented on some other providers?

src/parsec_client/operations/prepare_key_attestation.md
```diff
@@ -0,0 +1,40 @@
+# PrepareKeyAttestationParams
```
hug-dev, line 1: Should it be `PrepareKeyAttestation`?

src/parsec_client/operations/prepare_key_attestation_params.md
```diff
@@ -0,0 +1,56 @@
...
+
+| Name                 | Type   | Description                          |
+|----------------------|--------|--------------------------------------|
+| `attested_key_name`  | String | Name of the key to be attested       |
+| `attesting_key_name` | String | Name of the key to use for attesting |
+
+- if `attesting_key_name` is empty, the default key for the `ActivateCredential` mechanism will be
+   used
+
+## PrepareKeyAttestationOutput
```
hug-dev, line 43: I guess there should be 3 `#` here as it is also under `ActivateCredential`?

src/parsec_client/operations/attest_key_params.md
```diff
@@ -0,0 +1,66 @@
...
+`AttestationOutput`.
+
+**(EXPERIMENTAL) The parameters for key attestation are in an experimental phase. No guarantees are
+offered around the stability of the interface for any key attestation mechanism.**
+
+### Activate Credential (TPM provider) {#activatecredential}
```
hug-dev, line 13: I think you can remove one `#` here and to the sub titles to that section

hug-dev (approved):

paulhowardarm (approved):
Thanks for the follow-up changes - this looks good to me now.

hug-dev (approved):
🥅

PR #133, opened 2021-09-24
Add installation steps for openSUSE and SUSE

ggardet:


anta5010 (approved):

src/getting_started/opensuse.md
```diff
@@ -0,0 +1,81 @@
...
+
+A default config file is installed as `/etc/config.toml`. Only Mbed TLS (software) backend is enabled by default, but you can easily enable other backends, such as Tpm, CryptoAuthLib or Pkcs11.
+A full template is available at `/usr/share/doc/packages/config.toml`.
+
+You need to add the current linux user to `parsec-clients` group from YaST or from command line:
+```
+$ sudo usermod -a -G parsec-clients $USER
+```
+
+To apply the changes, you need to (re)login as `$USER`, or use `newgrp parsec-clients` command line
```
hug-dev, line 28: This shouldn't be needed anymore with the default authentication method as `UnixPeerCredentials`. This is only needed with the `Direct` method which should not really be used in the context of openSUSE I think (as Unix Domain socket are available and easy to use).

hug-dev (commented):
Nice, that's amazing! Shows that it's really easy to get started when the packages are there 😎 

You can ignore the link checker CI test but you would need to reformate your code using `vmdfmt`. There are some instructions in the [README](https://github.com/parallaxsecond/parsec-book#using).

src/getting_started/opensuse.md
ggardet: This is needed because `/var/run/parsec/` is owned by `parsec:parsec-clients`:
```
drwxr-x---  2 parsec parsec-clients   60 Sep 24 13:03 parsec
```
Not sure if it is desirable that all users can access `parsec.sock` by default.

hug-dev: I see!

> Not sure if it is desirable that all users can access parsec.sock by default.

In terms of Parsec threat model alone, when the unix peer credentials authenticator is used then it's perfectly fine (with the assumptions described in the TM related with OS security) to leave the socket open to all (it shouldn't later be changed to `Direct` but that can be prevented if the `Direct` one is compiled out).
With that authenticator, users will only be able to see/use keys created by them, security is based on Linux UID partitioning.

ggardet: Could we leave it as is for now? It will give me some time to update the RPM packaging properly and update packages accordingly in all openSUSE flavors.

hug-dev: Yes sure!

hug-dev (approved):
Formatting was a breeze 😎

PR #149, opened 2022-09-01
Add validation column for threat models

gowthamsk-arm:


ionut-arm (commented):
What's the difference between `N/A` and `-`?

gowthamsk-arm:
`N/A` is for test cases that are out of scope for Parsec testing. We have tickets created for tests with `-`. These will be added in future and updated in the book.

ionut-arm:
> We have tickets created for tests with -

Could you use a link to those tickets instead of `-`?

gowthamsk-arm:
Not required.

PR #151, opened 2022-10-17
Enhance and re-organise the Getting Started material

paulhowardarm:
The openSUSE and Linux/x86 Quickstart guides have been merged into one new overall installation guide including instructions for Fedora, Yocto and building from source.

Guidelines on `parsec-tool` usage also given with examples.

The README is reworked more along the lines of an FAQ, with pointers into these new docs, but also to relevant sections elsewhere in the book.

Signed-off-by: Paul Howard <paul.howard@arm.com>


paulhowardarm:
Link check failures appear to be API rate limiting issues rather than any actual problems with the links.

ionut-arm (approved):
Excellent, thank you for the great uplift!

src/getting_started/installation_options.md
```diff
@@ -0,0 +1,272 @@
...
+```
+export PARSEC_SERVICE_ENDPOINT="unix:$(pwd)/parsec.sock"
+```
+
+If parsec-tool is not installed into a directory included in `PATH`, then you also need to define
+`PARSEC_TOOL` environment variable with a full path to it:
+
+```
+export PARSEC_TOOL="$(pwd)/parsec-tool"
+```
```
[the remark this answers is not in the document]
anta5010: parsec-cli-tests.sh uses default value for and looking for  parsec-tool and openssl in PATH. These two (and OPENSSL) exports are required only in non-default cases.
I haven't checked fedora and opensuse packages buyt I would expect that they install and configure parsec and parsec-tool in default way

src/getting_started/README.md
```diff
@@ -1,13 +1,70 @@
...
+back-end, the [command-line tooling guide](parsec_tool_use.md) has some steps to help you check
+this.
+
+## I Want to Get Involved With the Parsec Community
+
+We will be delighted to have you! Head over to our
+[community](https://github.com/parallaxsecond/community) repository to discover how to join and
+contribute to the Parsec conversation on Slack, Zoom and GitHub.
 
 *Copyright 2021 Contributors to the Parsec project.*
```
gowthamsk-arm, line 70: Nit: As we have modified this file, should we also update copyright to 2022?

src/getting_started/opensuse.md
```diff
@@ -1,82 +1,6 @@
...
-
-If you make any change to its configuration, reload it with:
-
-```
-$ sudo systemctl restart parsec.service
-```
+This content has moved to the [installation
+guide](installation_options.md#option-1-install-parsec-using-the-package-manager).
 
 *Copyright 2021 Contributors to the Parsec project.*
```
gowthamsk-arm, line 6: Nit: Copyright updated to 2022?

src/getting_started/installation_options.md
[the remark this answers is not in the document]
ionut-arm: Correct, but the quickstart package is intended as a self-standing blob (hence it packages both service and tool), so you can assume that the `parsec_tool` is _here_, and that the socket will also be in the current directory. We don't introduce those exports in the "generic" CLI test script, but we could dump them in before packaging the quickstart package is my point.

src/getting_started/README.md
```diff
@@ -1,13 +1,70 @@
...
-- [Quickstart for Linux on x86](linux_x86.md): running Parsec with Mbed Crypto natively on your
-   machine, the fastest
-- [Quickstart for openSUSE and SUSE](opensuse.md): running Parsec with the package from openSUSE and
-   SUSE distributions
-- [Quickstart using a Docker container](docker.md): running a Docker container with all providers
-   included, the most complete
+Welcome to the Parsec Getting Started guide. Take a look through the headings below and follow the
+links that best describe what you would like to do.
+
+## I Want to Check if Parsec is Running
```
mohamedasaker-arm, line 8: Minor: It's more of a personal preference (which you can ignore)
Would it be okay if `I Want to` is removed

src/getting_started/installation_options.md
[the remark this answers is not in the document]
anta5010: Ah, I never liked that the quick start package uses non-default values.
But, AFAIU, this installation_options.md covers all installation cases not the quick start package only. So, it correctly explains what exports should be used in non-default cases.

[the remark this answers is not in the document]
ionut-arm: > But, AFAIU, this installation_options.md covers all installation cases not the quick start package only.

The whole file yes, but this specific section is under `Option 2: Download a Quick-Start Release`. Anyway, it was mostly a suggestion, don't think two exports add that much drag to the experience.

> Ah, I never liked that the quick start package uses non-default values.

Agree, but otherwise it would become quite an involved process of setting things up I think.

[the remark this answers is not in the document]
mohamedasaker-arm: @ionut-arm, there is no particular reason except maybe taking the script as is. However, since the package is for experimentation only and wants to deliver the smoothest experience. Then we might change the defaults for the `packaged` `parsec-cli-tests.sh` 
But again, as this export 
```export PARSEC_SERVICE_ENDPOINT="unix:$(pwd)/parsec.sock"``` 
is not avoidable if the user uses `parsec-tool` alone.

**Review by gowthamsk-arm — APPROVED** [review]

File: src/getting_started/README.md (hunk 2 above)
Comment (paulhowardarm) [reply src/getting_started/README.md:70]: I don't think we generally do that - we usually use the date of first publication for new files.

File: src/getting_started/installation_options.md [hunk 5]
```diff
@@ -0,0 +1,272 @@
...
+define `OPENSSL` environment variable with a full path to it:
+
+```
+export OPENSSL="<full path>/openssl"
+```
+
+## Option 3: Build from Source Code
+
+If the above installation options are not suitable, please refer to the instructions for [building
+from source code](../parsec_service/build_run.md) and [secure installation on
```
mohamedasaker-arm, line 258: I would assume that the line breaks after `on` (might be) due to the format restrictions.
However, it feels weird. If the format restrictions are the reason, we might need to revise them.
It's happening in different places, like line `261` and line 269.
ps: (worth sharing): I use vscode to view the output while editing (reviewing). It makes it easier to visualize the final shape

src/getting_started/parsec_tool_use.md
```diff
@@ -0,0 +1,536 @@
...
+
+```
+ID: 0x05 (CryptoAuthentication Library provider)
+Description: User space hardware provider, utilizing MicrochipTech CryptoAuthentication Library for ATECCx08 chips
+Version: 0.1.0
+Vendor: Arm
+UUID: b8ba81e2-e9f7-4bdd-b096-a29d0019960c
+```
+
+For Parsec to integrate with a [PSA
```
mohamedasaker-arm, line 194: Format: Sudden breaklines

mohamedasaker-arm (changes requested):

src/getting_started/installation_options.md
paulhowardarm: The text needs to be passed through vmdfmt with the same formatting rules as the CI checks, otherwise the CI fails.

src/getting_started/README.md
paulhowardarm: I worded it this way so that people could identify with the section(s) that are relevant to them.

src/getting_started/installation_options.md
paulhowardarm: The important thing is whether it renders correctly. If there is a problem with the rendering then we need to fix it, otherwise we have to do whatever the CI `vmdfmt` settings require,

src/getting_started/parsec_tool_use.md
paulhowardarm: Again, I think this is just `vmdfmt` artefacts, but the CI will fail if we structure it differently. The important thing is whether it renders correctly.

PR #157, opened 2023-03-07
Add release checklist and testing information

gowthamsk-arm:
This commit adds the checklist containing specific tasks to be completed in stage of the release and the testing strategy which defines the hardware/backends that are used to for running the tests.

### DIFF
--- a/src/contributing/release_checklist.md
+++ b/src/contributing/release_checklist.md
@@ -1,0 +1,54 @@
+# The Parsec Release Checklist
+
+The release process is divided into 4 stages as shown below. Each stage has a discrete set of
+activities to be completed before progressing to the next stage. The list shown below is only a
+template, as there might be additional tasks specific to that release that might be added to
+different stages.
+
+Stages:
+
+- Prior to Code freeze
+   - [ ]  Publish new versions of all managed crates and configure parsec service and tool to use
+      them. See
+      this [guide](https://parallaxsecond.github.io/parsec-book/contributing/package_management.html).
+   - [ ]  Fix any pending issues
+   - [ ]  Check common dependency crates versions between the parsec service, parsec tool, and
+      other crates
+   - [ ]  Review PRs across Parsec repos
+   - [ ]  Merge approved PRs
+   - [ ]  Check the milestone list
+- During code freeze
+   - [ ]  Issue release candidate tag x.x.x-rc1
+   - [ ]  OpenSUSE packaging for the RC
+   - [ ]  Pull in the latest git dependencies in the crates to ensure the latest release would
+      work.
+   - [ ]  Execute the normal e2e tests on Raspberry Pi hardware:
+      - mbed-crypto
+      - TPM
+      - Nitrokey HSM (pluggable USB)
+   - [ ]  Execute fuzz testing for a week
+   - [ ]  Ask the Linux distributions packagers to try the tagged version
+   - [ ]  Ask the Parsec Yocto maintainers to try the tagged version
+   - [ ]  Ask any interested people to try the tagged version
+   - [ ]  Review the book and make sure all pages are up-to-date. In particular, make sure the
+      Threat Model is up to date.
+   - [ ]  Make sure all the markdown pages (like the README) look good.
+   - [ ]  Make sure the rustdoc pages (on docs.rs) are looking good (before and after the release).
+   - [ ]  If any bugs are reported, we fix bugs and then issue another release candidate, x.x.x-rcx
+   - [ ]  When All reported issues are addressed, make a GitHub release out of the last tag, which
+      includes its changelog, and ask maintainers to release this version of Parsec.
+- Release
+   - [ ]  Push the release tag for the service and the parsec-tool
+   - [ ]  Publish the new quickstart bundle and update the quickstart guide with it
+- Post-release
+   - [ ]  Close issues fixed in the release and the past milestone
+   - [ ]  Update, test, and upstream Yocto recipes for Parsec and parsec-tool
+   - [ ]  Make a new issue milestone for the next release
+   - [ ]  Update the [Roadmap](https://github.com/parallaxsecond/community/blob/main/ROADMAP.md)
+   - [ ]  Update the [Parsec Release
+      Process](https://parallaxsecond.github.io/parsec-book/contributing/release_process.html) 
+      with what we learned.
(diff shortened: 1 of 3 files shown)


src/contributing/release_process.md
```diff
@@ -127,6 +127,47 @@ perform whatever testing is necessary to verify that the new feature has reached
...
+   - import_ecc_key
+   - import_ecc_key_twice
+- TPM: 5 failures
+   - simple_asym_encrypt_rsa_pkcs
+   - activate_credential_bad_data
+   - activate_credential_ecc
+   - activate_credential_rsa
+   - check_name
+
+The new release candidate should ideally not introduce any regressions.
```
paulhowardarm, line 169: Language feels a bit vague here: "ideally not introduce any regressions". What do we really mean, and what is the process? Let's try to tighten this up. Perhaps we need to say that any regressions will result in the creation of a new GH issue and will be flagged to the community for assessment as to whether it is a blocking issue for the release, during the RC phase?

src/contributing/release_process.md
```diff
@@ -127,6 +127,47 @@ perform whatever testing is necessary to verify that the new feature has reached
...
+
+- mbed crypto
+- TPM
+- Nitrokey HSM
+
+Further platforms and backends will be added to this in future releases. The results of the release
+testing are critical in determining if the parsec service and the tool is ready for release, and
+whether it meets the requirements and expectations of the end users.
+
+The following failures are expected in the RaspberryPi platform for the specific backends.
```
paulhowardarm, line 141: Do we understand the failures? Are they actual Parsec defects or specific interop issues with the NitroKey HSM? If the failures are deemed acceptable, do we have a way to subset the e2e tests such that they are excluded from the release checking? I see it as a problem that we have a cascade of "known" failures, because it puts the onus on the engineer to manually inspect for regressions. I appreciate that this is not so much a code review comment, but rather an observation about the process itself. For the time being, if we have to document it this way, I'm okay with it in terms of this PR.

src/contributing/release_process.md
```diff
@@ -127,6 +127,47 @@ perform whatever testing is necessary to verify that the new feature has reached
...
+   - activate_credential_bad_data
+   - activate_credential_ecc
+   - activate_credential_rsa
+   - check_name
+
+The new release candidate should ideally not introduce any regressions.
+
 ## Release Expectations
 
 There are currently no formal release expectations documented for the Parsec project. Release
```
paulhowardarm, line 173: Are we in a position to put in a few basic release expectations at this point? If we follow the checklist, then we implicitly have some expectations about what will work, and what platforms we've tested on. At a minimum, perhaps just reword this section to say that formal release expectations are still TBD, but we can at least reference the checklist and set the expectation that the checklist has been followed, and therefore any new Parsec release can at least be assumed to work on a fixed set of platforms. This would set us up nicely for enhancing that list in the future.

paulhowardarm (changes requested):
Thanks for the PR. Looks good overall. Nice enhancement to the process. I have some suggestions.

src/contributing/release_process.md
gowthamsk-arm: Yes, I agree with your point. That should be clear as to what we need to do in case of any regressions. Will update this line with your suggestion.

src/contributing/release_process.md
gowthamsk-arm: Yeah, considering the checklist we can put up some basic release expectations covering the platforms/backends tested and the OpenSUSE distro. Will update this part.

src/contributing/release_checklist.md
```diff
@@ -0,0 +1,54 @@
+# The Parsec Release Checklist
+
+The release process is divided into 4 stages as shown below. Each stage has a discrete set of
+activities to be completed before progressing to the next stage. The list shown below is only a
+template, as there might be additional tasks specific to that release that might be added to
+different stages.
+
+Stages:
+
+- Prior Code freeze
```
tgonzalezorlandoarm, line 10: "Prior to" ?

src/contributing/release_checklist.md
```diff
@@ -0,0 +1,54 @@
...
+   - [ ]  Publish new versions of all managed crates and configure parsec service and tool to use
+      them. See
+      this [guide](https://parallaxsecond.github.io/parsec-book/contributing/package_management.html).
+   - [ ]  Fix any pending issues
+   - [ ]  Check common dependency crates versions between the parsec service, parsec tool, and
+      other crates
+   - [ ]  Review PRs across Parsec repos
+   - [ ]  Merge PRs got approved
```
tgonzalezorlandoarm, line 18: "Merge approved PRs" ?

src/contributing/release_checklist.md
```diff
@@ -0,0 +1,54 @@
...
+- During code freeze
+   - [ ]  Issue release candidate tag x.x.x-rc1
+   - [ ]  OpenSUSE packaging for the RC
+   - [ ]  Pull in the latest git dependencies in the crates to ensure the latest release would
+      work.
+   - [ ]  Execute the normal e2e tests on Raspberry Pi hardware:
+      - mbed-crypto
+      - TPM
+      - Nitrokey HSM (pluggable USB)
+   - [ ]  Execute fuzz testing for some time
```
tgonzalezorlandoarm, line 29: Could we maybe expand on the "for some time"?

gowthamsk-arm: Yes, will update it.

src/contributing/release_process.md
gowthamsk-arm: Created #676 and #677 for the detailed investigation before we add this to the parsec book.

src/contributing/release_process.md
```diff
@@ -127,16 +127,28 @@ perform whatever testing is necessary to verify that the new feature has reached
...
 
+As part of the release testing, we run the end-to-end tests on a RaspberryPi for the following
+backends:
+
+- mbed crypto
+- TPM
+- Nitrokey HSM
+
+Further platforms and backends will be added to this in future releases. The results of the release
+testing are critical in determining if the parsec service and the tool is ready for release, and
```
tgonzalezorlandoarm, line 138: are ready for release, and whether they meet

tgonzalezorlandoarm (approved):

paulhowardarm (approved):
Thank you, this looks good.

PR #158, opened 2023-04-04
Update quickstart info and release version

gowthamsk-arm:
This commit contains the following changes:
- The quickstart pages just have links to the installation guide. These are removed and installation page is renamed.
- References to 1.1.0 release is changed to 1.2.0

### DIFF
--- a/src/getting_started/docker.md
+++ b/src/getting_started/docker.md
@@ -1,5 +1,6 @@
 # Docker
 
-TODO
+Please refer to the [installation
+guide](installation_options.md#option-3-use-a-quickstart-docker-image).
 
-*Copyright 2021 Contributors to the Parsec project.*
+*Copyright 2021-2023 Contributors to the Parsec project.*
--- a/src/getting_started/installation_options.md
+++ b/src/getting_started/installation_options.md
@@ -173,7 +173,7 @@ export PATH=${PATH}:$(pwd)/../bin
 Start the Parsec service with this command:
 
 ```
-./parsec &
+parsec &
 ```
 
 You should see some lines of console output as the service starts, ending with the following:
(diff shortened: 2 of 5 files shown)


anta5010 (approved):
LGTM

paulhowardarm (commented):
Are we confident that deleting these files will not result in broken links? Our previous strategy was to maintain the older Quickstart files with a "content has moved" message, just in case there were any specific links to those pages. The Docker one was never populated to begin with, but arguably it should be treated in a uniform manner (with a similar "content has moved" message).

There might have been a community discussion where we talked about removing those files, but my memory is hazy. Did we conclude that removing them was the right thing to do? If so, then that's fine.

adamjohnstonarm:
> Within the parsec-book repo, I've made sure there are no references to the deleted pages and I can't recollect the community discussion related to this. I'm happy to stick with the previous strategy.
> 
> As a reader, those pages looked odd while reviewing the book so that prompted me to make this change.



> Are we confident that deleting these files will not result in broken links? Our previous strategy was to maintain the older Quickstart files with a "content has moved" message, just in case there were any specific links to those pages. The Docker one was never populated to begin with, but arguably it should be treated in a uniform manner (with a similar "content has moved" message).
> 
> There might have been a community discussion where we talked about removing those files, but my memory is hazy. Did we conclude that removing them was the right thing to do? If so, then that's fine.

If I understood the original changes, we were keeping the pages but removing the links to them from the ToC. That should be ok.

gowthamsk-arm:
In the first version of my changes I was deleting the pages as well.

gowthamsk-arm:
> I see that now. Unless SUMMARY.md is automatically generated or it prunes un-referenced content, it should still be possible to leave the stub/redirect pages intact, but also remove the links from the Table of Contents. 

https://rust-lang.github.io/mdBook/cli/build.html 
If a file is not mentioned in Summary.md, it won't be considered for the build. 

Will update the docker link

adamjohnstonarm (approved):
LGTM

paulhowardarm (approved):

PR #164, opened 2023-10-24
Change parsec version to 1.3.0

tgonzalezorlandoarm:
Signed-off-by: Tomás González <tomasagustin.gonzalezorlando@arm.com>

### DIFF
--- a/src/getting_started/installation_options.md
+++ b/src/getting_started/installation_options.md
@@ -241,13 +241,14 @@ export PARSEC_SERVICE_ENDPOINT="unix:$(pwd)/parsec.sock"
 ```
 
 If parsec-tool is not installed into a directory included in `PATH`, then you also need to define
-`PARSEC_TOOL` environment variable with a full path to it:
+the `PARSEC_TOOL` environment variable with a full path to it. Assuming the current working
+directory is `quickstart-1.2.0-linux_x86_64/quickstart`:
 
 ```
-export PARSEC_TOOL="$(pwd)/parsec-tool"
+export PARSEC_TOOL="$(pwd)/../bin/parsec-tool"
 ```
 
-To run the script, simply execute it without any arguments as follows:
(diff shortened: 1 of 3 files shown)


src/getting_started/installation_options.md
```diff
@@ -241,13 +241,15 @@ export PARSEC_SERVICE_ENDPOINT="unix:$(pwd)/parsec.sock"
 ```
 
 If parsec-tool is not installed into a directory included in `PATH`, then you also need to define
-`PARSEC_TOOL` environment variable with a full path to it:
+`PARSEC_TOOL` environment variable with a full path to it. Assuming the current working directory is
+the root folder of the `parsec-tool`:
 
 ```
-export PARSEC_TOOL="$(pwd)/parsec-tool"
+export PARSEC_TOOL="$(pwd)/target/debug/parsec-tool"
```
gowthamsk-arm, line 248: I think the context is using the quick start bundle where we have the built binaries. 
So this change is not needed as it should be straight forward.

src/getting_started/installation_options.md
```diff
@@ -241,13 +241,15 @@ export PARSEC_SERVICE_ENDPOINT="unix:$(pwd)/parsec.sock"
...
 ```
 
-To run the script, simply execute it without any arguments as follows:
+To run the script, simply go to `quickstart-1.2.0-linux_x86_64/quickstart` and execute it without
+any arguments as follows:
```
gowthamsk-arm, line 252: Same as above.

src/getting_started/installation_options.md
tgonzalezorlandoarm: The context here is:
`If parsec-tool is not installed into a directory included in PATH, then you also need to define PARSEC_TOOL environment variable with a full path to it:`

So that means that this is used to indicate how to do it if you are building or installing into a different path, no?

gowthamsk-arm (approved):

